PR #20, opened 2020-03-11
Component: Transaction Tags

aulneau:
Resolves: https://github.com/blockstack/explorer/issues/11

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -1,5 +1,15 @@
 import React from 'react';
+import { Tag, Transaction } from '@components/tags';
+import { Flex, Stack } from '@blockstack/ui';
 
-const Home: React.FC = () => <>Home</>;
+const Home: React.FC = () => (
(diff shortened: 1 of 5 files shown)


src/pages/index.tsx
```diff
@@ -1,5 +1,21 @@
...
 
-const Home: React.FC = () => <>Home</>;
+// @ts-ignore
+const Home: React.FC = () => (
+  <Flex height="100vh" width="100%" align="center" justify="center">
+    <Stack spacing="loose">
+      {Object.keys(Transaction).map(label => (
+        <Tag
+          type={
+            // @ts-ignore
```
kyranjamie: `<Tag type={Transaction[label as keyof typeof Transaction]} />`

// @ts-ignore is big boo-boo in TypeScript land

aulneau: yeah but this is just to display them :) this won't happen in real life

aulneau: will update though!

PR #21, opened 2020-03-12
Component: Status

aulneau:


src/components/status.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import { Flex, Text, Box, Spinner, BoxProps } from '@blockstack/ui';
+import { CheckmarkCircleIcon, ExclamationMarkCircleIcon } from '@components/svg';
+
+export enum Statuses {
```
kyranjamie: With enums I often find the resulting API to make more sense when using singular naming

aulneau: Yeah, just a conflict with the component named Status

PR #23, opened 2020-03-18
feat: initial data fetching

kyranjamie:
To run this PR, you must first boot the mock server from the [`explorer-api`](https://github.com/blockstack/blockstack-explorer-api) repo, 

**Run in that repo**
```
yarn mock
```


src/common/api/transactions.tsx
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios';
```
aulneau: Personally I am not a huge fan of using axios -- I prefer fetch/unfetch -> https://github.com/developit/unfetch/tree/master/packages/isomorphic-unfetch

kyranjamie: Can you elaborate why? Fetch is frustrating as we need to call the `json()` method, and we always want json.

4.6k stars versus 70k?

aulneau: Sure, of course. All your points are valid.

I think the biggest thing is if it's just to remove the use of `.json()`, typically I have in the past created a wrapper function for data fetches, using fetch. Additionally we will probably have a method for abstracting out the url anyways, so it would make sense to have the extra line of '.json()` included in this method.

There is a significant size difference between the two. 4.4kb vs 582b. I'd understand using axios if we would take advantage of any of it's other features, but outside of only fetching, there is no reason I see to why it's advantageous over polyfilling the standard fetch library. It's like using moment for time management for one method, when there are so many other, smaller libs.

aulneau: > The latest commit of this library breaks their CI 
also I don't think this is true, it seems to be an error on the view that I linked, `master` is passing fine. The package is maintained and created by the creator of preact, so imho we don't really have to worry about if it's popular or not.

kyranjamie: For argument's sake, I've made this change. 

My approach would be different here, though. An essential tool building a web app is a decent HTTP client. Wrapper libs like axios are popular because fetch alone isn't sufficient. This request to use isomorphic-unfetch, over axios for 4kB, is dangerously close to being a premature optimisation. I'd rather remove axios for this library when we're struggling with a large bundle, than before we've built anything.

PR #24, opened 2020-03-23
feat: transaction page implemented

aulneau:


src/components/rows.tsx
```diff
@@ -0,0 +1,162 @@
...
+    currency: string;
+  };
+  block: number;
+  post_conditions: string;
+  sponsored: boolean;
+  state: string;
+  timestamp: number;
+}
+
+interface TxSchema {
```
kyranjamie: repeated type

src/components/rows.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { Caption } from '@components/typography';
+import { Card } from '@components/card';
+import { toSnakeCase } from '@common/utils';
+
+interface TxSchema {
+  txid: string;
+  sender: string;
+  recipient: string;
+  type: number;
+  fees: {
```
kyranjamie: Better to make a separate type for Fees above, so it doesn't need to be repeated on line :44

kyranjamie:
@zone117xs PR on sidecar returns data we can hook up now, so will be great to get this PR merged, even if not totally ready

src/components/rows.tsx
kyranjamie: These can go now, see models folder

aulneau:
Rodger, sounds good. I will push and fix any conflicts, and we can merge today @kyranjamie

PR #43, opened 2020-03-25
feat: scaffold base homepage functionality

kyranjamie:
image

I've added the `store` npm package, rather than attempting to use redux-persist. It's a < 1kB package that wraps to string/to object functionality, makes for much more readable code.

Looking for feedback on the hook, this is the first one I've made.

### DIFF
--- a/src/common/hooks/use-recently-viewed-tx.tsx
+++ b/src/common/hooks/use-recently-viewed-tx.tsx
@@ -1,0 +1,21 @@
+import { useState, useEffect } from 'react';
+import { store } from '@common/utils';
+
+const storageKey = 'recentlyViewedTx';
+
+export const useRecentlyViewedTx = (txid?: string) => {
+  if (!process.browser) return;
+
+  const storedTxs = store.get(storageKey, []);
+  const [viewedTransactions, setViewedTransactions] = useState(storedTxs);
+
+  useEffect(() => {
+    if (txid) {
+      const newTxList = [...viewedTransactions, txid];
+      setViewedTransactions(newTxList);
+      store.set(storageKey, newTxList);
+    }
+  }, [txid]);
+
+  return viewedTransactions;
+};
--- a/src/common/hooks/use-recently-viewed-tx.tsx
+++ b/src/common/hooks/use-recently-viewed-tx.tsx
@@ -1,5 +1,5 @@
 import { useState, useEffect } from 'react';
-import { store } from '@common/utils';
+import { store, dedupe } from '@common/utils';
 
 const storageKey = 'recentlyViewedTx';
 
@@ -11,7 +11,7 @@ export const useRecentlyViewedTx = (txid?: string) => {
 
   useEffect(() => {
     if (txid) {
-      const newTxList = [...viewedTransactions, txid];
+      const newTxList = dedupe([...viewedTransactions, txid]);
       setViewedTransactions(newTxList);
       store.set(storageKey, newTxList);
     }
--- a/src/components/dev-links.tsx
+++ b/src/components/dev-links.tsx
@@ -1,0 +1,9 @@
+import React from 'react';
+import Link from 'next/link';
+import { Box } from '@blockstack/ui';
+
+export const DevLinks = () => (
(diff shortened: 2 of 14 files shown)


src/common/hooks/use-recently-viewed.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const storedTxs = store.get(storageKey, []);
+  const [viewedTransactions, setViewedTransactions] = useState(storedTxs);
+
+  useEffect(() => {
+    if (txid) {
+      const newTxList = [...viewedTransactions, txid];
+      setViewedTransactions(newTxList);
+      store.set(storageKey, newTxList);
+    }
+  }, [txid]);
```
aulneau: We aren't doing anything to filter duplicates. Might be better to check if exists, and push to the end of the array instead of adding a dupe.

src/pages/index.tsx
```diff
@@ -1,17 +1,45 @@
...
+    e.preventDefault();
+    if (!query.trim().length) return;
+    await Router.push('/txid/' + query);
+  };
+
+  const recentlyViewedTransactions = useRecentlyViewed();
+
+  return (
+    <Page>
+      {JSON.stringify(recentlyViewedTransactions, null, 2)}
```
aulneau: Do we still want this here?

src/pages/index.tsx
```diff
@@ -1,17 +1,45 @@
...
+      <Flex
+        as="form"
+        flexDirection="column"
+        align="center"
+        maxWidth="544px"
+        mt="20vh"
+        justify="center"
+        onSubmit={handleSubmit}
+      >
+        <Text as="h1" fontSize="36px" display="block" width="100%" textAlign={['center', 'left']} mb={6}>
```
aulneau: It would be good to start using the named spacing units we've added, in place of numbers for margin/padding: https://github.com/blockstack/ux/blob/master/packages/ui/src/theme/sizes.tsx#L30

src/components/dev-links.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import Link from 'next/link';
+import { Box } from '@blockstack/ui';
+
+export const DevLinks = () => (
+  <Box position="absolute" bottom={4} right={4}>
```
aulneau: Please use named spacing units: https://github.com/blockstack/ux/blob/master/packages/ui/src/theme/sizes.tsx#L30

src/pages/txid/[txid].tsx
```diff
@@ -1,22 +1,32 @@
 import React from 'react';
+import Link from 'next/link';
 import { useSelector } from 'react-redux';
 
 import { RootState } from '@store';
 import { fetchTransactionDone, selectTransaction } from '@store/transactions';
 import { ReduxNextPageContext } from '@common/types/next-store';
 import { fetchTx } from '@common/api/transactions';
+import { useRecentlyViewed } from '../../common/hooks/use-recently-viewed';
```
Comment (aulneau) [inline src/pages/txid/[txid].tsx]: Should be `@common` instead of `../../common`

src/common/hooks/use-recently-viewed.tsx
aulneau: I'd also maybe rename it to `use-recently-viewed-tx`/`useRecentlyViewedTx`

kyranjamie: Yep, good spot. This isn't quite final yet as will need to save more than just txid, per designs.

src/common/hooks/use-recently-viewed-tx.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const useRecentlyViewedTx = (txid?: string) => {
+  if (!process.browser) return;
+
+  const storedTxs = store.get(storageKey, []);
+  const [viewedTransactions, setViewedTransactions] = useState(storedTxs);
+
+  useEffect(() => {
+    if (txid) {
+      const newTxList = [...viewedTransactions, txid];
```
aulneau: Easy solution for now: 

```suggestion
      const newTxList = [...new Set([...viewedTransactions, txid])];
```

PR #44, opened 2020-03-30
feat: add recently viewed component

kyranjamie:
image

### DIFF
--- a/src/common/hooks/use-recently-viewed-tx.tsx
+++ b/src/common/hooks/use-recently-viewed-tx.tsx
@@ -1,21 +1,28 @@
 import { useState, useEffect } from 'react';
-import { store, dedupe } from '@common/utils';
+import { store } from '@common/utils';
+import { Transaction } from '@models/transaction.interface';
 
 const storageKey = 'recentlyViewedTx';
 
-export const useRecentlyViewedTx = (txid?: string) => {
-  if (!process.browser) return;
+interface CachedTx extends Transaction {
+  viewedDate: string;
+}
+
+export const useRecentlyViewedTx = (transaction?: Transaction) => {
+  if (!process.browser) return [];
 
   const storedTxs = store.get(storageKey, []);
-  const [viewedTransactions, setViewedTransactions] = useState(storedTxs);
+  const [viewedTransactions, setViewedTransactions] = useState<Transaction[]>(storedTxs);
 
   useEffect(() => {
-    if (txid) {
-      const newTxList = dedupe([...viewedTransactions, txid]);
+    if (transaction) {
+      const newTx: CachedTx = { ...transaction, viewedDate: new Date().toISOString() };
+      const cachedTxs = viewedTransactions.filter(({ tx_id }) => tx_id !== transaction.tx_id);
+      const newTxList = [...cachedTxs, newTx];
       setViewedTransactions(newTxList);
       store.set(storageKey, newTxList);
     }
-  }, [txid]);
+  }, [transaction?.tx_id]);
 
   return viewedTransactions;
 };
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -19,3 +19,9 @@ export const toSnakeCase = (str: string) => {
 
   return string.toLowerCase();
 };
+
+export const truncateMiddle = (input: string, offset: number) => {
+  const start = input.substr(0, offset);
+  const end = input.substr(input.length - offset, input.length);
+  return `${start}…${end}`;
+};
--- a/src/components/home-nav.tsx
+++ b/src/components/home-nav.tsx
@@ -1,11 +1,16 @@
 import React from 'react';
 import { Flex, Box } from '@blockstack/ui';
 import { BlockstackLogo } from './icons/blockstack-logo';
+import Link from 'next/link';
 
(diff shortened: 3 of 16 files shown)


src/common/hooks/use-recently-viewed-tx.tsx
```diff
@@ -1,21 +1,28 @@
...
-      const newTxList = dedupe([...viewedTransactions, txid]);
+    if (transaction) {
+      const newTx: CachedTx = { ...transaction, viewedDate: new Date().toISOString() };
+      const cachedTxs = viewedTransactions.filter(({ tx_id }) => tx_id !== transaction.tx_id);
+      const newTxList = [...cachedTxs, newTx];
       setViewedTransactions(newTxList);
       store.set(storageKey, newTxList);
     }
-  }, [txid]);
+  }, [transaction?.tx_id]);
```
aulneau, line 25: I wonder, do we need to put process.browser as a dep in the effect, so that the component will rerender on hydration and get the cached tx's?

src/common/utils.ts
```diff
@@ -4,3 +4,9 @@ import lclStorage from 'store/storages/localStorage';
 export const store = engine.createStore([lclStorage]);
 
 export const dedupe = (array: string[]) => [...new Set(array)];
+
+export const truncateMiddle = (input: string, offset: number) => {
+  const start = input.substr(0, offset);
+  const end = input.substr(input.length - offset, input.length);
+  return `${start}…${end}`;
```
aulneau, line 11: Should we use `&hellip;`?

src/components/home-nav.tsx
```diff
@@ -1,11 +1,14 @@
...
 import { BlockstackLogo } from './icons/blockstack-logo';
+import Link from 'next/link';
 
 export const HomeNav = () => (
   <Flex justifyContent="space-between" alignItems="center" height="64px">
     <Box>
-      <BlockstackLogo />
+      <Link href="/">
+        <BlockstackLogo />
+      </Link>
```
aulneau, line 11: I believe we have to wrap the child of link in an `a` tag

src/components/recently-viewed.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface RecentlyViewedListItemProps extends BoxProps {
+  transaction: Transaction;
+  isLast: boolean;
+}
+
+export const RecentlyViewedListItem = ({ transaction, isLast }: RecentlyViewedListItemProps) => {
+  const [isHovered, bind] = useHover();
+  return (
+    <Link href="/txid/[txid]" as={`/txid/${transaction.tx_id}`}>
```
aulneau, line 62: The flex should have `as="a"` and the Link should have `passHref`: https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

aulneau:
I think lint:fix needs to be ran, too @kyranjamie

src/common/hooks/use-recently-viewed-tx.tsx
kyranjamie: This uses localStorage, so we don't want it to run on the server.

src/common/utils.ts
kyranjamie: Not to the best of my knowledge, what's wrong with using the unicode character directly?

src/components/recently-viewed.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { Transaction } from '@models/transaction.interface';
+import Link from 'next/link';
+
+dayjs.extend(relativeTime);
+
+interface RecentlyViewedProps extends FlexProps {
+  transactions: Transaction[];
+}
+
+export const RecentlyViewed = ({ transactions, ...rest }: RecentlyViewedProps) => {
```
aulneau, line 18: There is a box-shadow applied to this in figma: `box-shadow: 0px 8px 16px rgba(27, 39, 51, 0.08);`

src/common/hooks/use-recently-viewed-tx.tsx
aulneau: Yeah, def, I suppose what I am asking when it's loaded on the client for the first time, does it rerender, or does it use the empty array? I could see a situation that would happen where it loads and until we pass a new txid, then it would get all of the cached items that it should have fetched at the beginning.

src/common/utils.ts
aulneau: I think if we use the unicode, it won't break the line if there is a situation where it would break: eg

```
asdas..
.asda
```

src/common/hooks/use-recently-viewed-tx.tsx
kyranjamie: I'm not sure I follow. This effect will only run when the first, and subsequent unique txs are viewed.

kyranjamie:
The linting seems to be correct to me. I can't see why Now is failing as I don't have permissions @aulneau

src/common/utils.ts
kyranjamie: I had a quick google but can't find anywhere suggesting you shouldn't use unicode characters in JSX. I thought the days of `&amp;` and `&hellip;` were past us.

PR #45, opened 2020-03-31
feat: hook up to api, include transaction interface

kyranjamie:
For time being, this type has been copied from the sidecar repo.

Going forward, I hope we can have all these types generated from the JSON schmas so there's no duplication.


agraebe:
is the preview useful for the transaction interface? I usually check it out but can't see the new interfaces (likely not linked from the main page?)

PR #57, opened 2020-04-09
refactor: integrate with sidecar

kyranjamie:
Takes Thomas' work and hooks it up to the sidecar and the types it outputs.


src/components/transaction-details.tsx
```diff
@@ -63,48 +39,55 @@ const transformDataToRowData = (d: Transaction) => {
...
+      copy: d.sender_address,
     },
     {
       label: {
         children: 'Called by',
       },
-      children: d.recipient,
-      copy: d.recipient,
+      children: d.sender_address,
+      copy: d.sender_address,
```
aulneau, line 57: Should these values be the same?

kyranjamie: They were before?

kyranjamie: ah see what you mean now

.eslintrc.js
```diff
@@ -1,73 +1,18 @@
...
-  extends: [
-    'plugin:@typescript-eslint/recommended',
-    'plugin:@typescript-eslint/recommended-requiring-type-checking',
-    'plugin:react/recommended',
-    'prettier/@typescript-eslint',
-    'plugin:jest/recommended',
-    'plugin:prettier/recommended'
-    // 'plugin:jsx-a11y/recommended'
-  ],
+  extends: ['@blockstack/eslint-config'],
```
aulneau, line 2: love this

PR #60, opened 2020-04-13
feat: add tx page header

kyranjamie:
image

### DIFF
--- a/src/components/search-bar.tsx
+++ b/src/components/search-bar.tsx
@@ -1,27 +1,39 @@
 import React from 'react';
-import { Box, Input } from '@blockstack/ui';
+import { Box, Input, BoxProps } from '@blockstack/ui';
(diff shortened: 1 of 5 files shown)


src/components/search-bar.tsx
```diff
@@ -1,27 +1,39 @@
...
       bottom={0}
       left={0}
       right={0}
-      pl="50px"
-      height="64px"
-      placeholder="Search for transactions or contracts"
-      type="text"
+      pl={inputOffset}
+      lineHeight="20px"
+      p={0}
```
aulneau, line 22: we have `pl` and then `p` which would overwrite `pl`. I'd put `pl` after `p`, or ```p={`0 0 0 ${inputOffset}`}```

kyranjamie: It doesn't overwrite it, but moved after anyway.

PR #65, opened 2020-04-15
Feat/add opengraph metadata

kyranjamie:


src/pages/_document.tsx
```diff
@@ -27,4 +30,31 @@ export default class MyDocument extends Document {
...
+    return (
+      <html {...this.helmetHtmlAttrComponents}>
+        <Head>{this.helmetHeadComponents}</Head>
+        <body {...this.helmetBodyAttrComponents}>
+          <Main />
+          <NextScript />
+        </body>
+      </html>
+    );
+  }
```
aulneau, line 59: next.js has a function for this built in, in the past that is exclusively what I've used: https://nextjs.org/docs/api-reference/next/head

What was the rationale for using react-helmet? it seems like an unneeded package when the functionality is built into next.js, it adds [16kb/5.5kb](https://bundlephobia.com/result?p=react-helmet@6.0.0) to the final bundle.

PR #68, opened 2020-04-16
feat: add footer link styles

kyranjamie:
image


src/components/footer-links.tsx
```diff
@@ -0,0 +1,40 @@
...
+);
+
+export const FooterLinks = () => (
+  <Flex flex={1} alignItems="flex-end" mb={['base', 'base-loose', 'loose']}>
+    <FooterLink
+      onClick={async () => {
+        const { results } = await fetchTxList();
+        console.log(results);
+        const randomTx = results[Math.floor(Math.random() * results.length)];
+        await Router.push(`/txid/${randomTx.tx_id}`);
```
aulneau, line 26: Anytime we are navigating internally we should do this approach: https://github.com/blockstack/explorer/blob/54eb7e3e2bed69a18d6487ae32d69abe476b7834/src/common/search.tsx#L15

When we just do `Router.push(path)`, it causes a hard load of the page, rather than doing client side navigation (keeping state, etc).

aulneau, line 26: Anytime we are navigating internally we should do this approach: https://github.com/blockstack/explorer/blob/54eb7e3e2bed69a18d6487ae32d69abe476b7834/src/common/search.tsx#L15

When we just do `Router.push(path)`, it causes a hard load of the page, rather than doing client side navigation (keeping state, etc).

PR #69, opened 2020-04-16
fix: add more details to opengraph

kyranjamie:


public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
zone117x, line 2: Out of curiosity, why this config?

kyranjamie: This is like the de facto allow everything robots.txt

kyranjamie: actually, it isn't. Good call, need to remove the `/`

PR #75, opened 2020-04-19
feat: search for tx by contract_id

aulneau:
### What this does

This PR implements the ability to search for a transaction by either the tx_id or the contract_id. Additionally it implements basic validation for the input: checks if it contains a `.`, and then validates the first half as a stacks transaction, and then a regex for the `contract-name`.

When the user navigates to that page, the url is updated with the correct tx_id, rather than keeping the contract name in the url.

contract-search


src/common/hooks/use-transaction-state.ts
```diff
@@ -2,17 +2,30 @@ import { useSelector } from 'react-redux';
...
+
+  if (specificTx) {
+    transaction = useSelector((state: RootState) => ({
+      transaction: selectTransaction(specificTx)(state),
+    })).transaction;
+  } else if (tx_id) {
+    transaction = useSelector((state: RootState) => ({
+      transaction: selectTransaction(tx_id)(state),
+    })).transaction;
+  }
```
kyranjamie, line 26: What's the difference between a `tx_id` and a `specificTx`?

src/common/utils.ts
```diff
@@ -1,14 +1,32 @@
...
+import { c32addressDecode } from 'c32check';
 
 export const store = engine.createStore([lclStorage]);
 
+/**
+ * validateStacksAddress
+ *
+ * @param {String} stacksAddress - the STX address to validate
+ */
+const validateStacksAddress = (stacksAddress: string) => {
```
kyranjamie, line 12: Wouldn't a `(stxAddress: string) => boolean` signature be more helpful here? Rather than true or throws error?

src/common/hooks/use-transaction-state.ts
aulneau: You can use this hook to fetch the most recent TX id searched for, or supply another txid.

src/common/utils.ts
aulneau: Yes, I just pulled this from stacks-utils, happy to update it

src/common/hooks/use-transaction-state.ts
kyranjamie: By fetch you mean, get from store, not make a request?

I just find it a little confusing to read, and afaict, breaks SRP. If we want to get the most recently searched for tx, couldn't we keep this hook as it was before,:
- Persist an array of queried txs
- Where needed, ` const id = selectMostRecentTxId(state); const tx = useTxState(id)`

aulneau: yeah, sure that makes sense! will update

PR #88, opened 2020-04-21
fix: ssr and style improvements for highlighter.tsx

aulneau:


src/components/highlighter/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  types: string[];
+  content: string;
+  empty?: boolean;
+};
+
+export type StyleObj = {
+  [key: string]: string | number | null;
+};
+
+export type TokenOutputProps = {
```
kyranjamie, line 12: Can be interfaces. Even though this works it's a much more common pattern to save `type` for when creating new types of existing interfaces. The TS docs uses them this way.

src/components/codeblock.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const CodeBlock = ({ code }: { code: string }) => {
+  const [editorCode] = useState(code?.toString().trim());
+
+  return (
+    <Box
+      // @ts-ignore
+      css={css({
+        '*': {
+          whiteSpace: 'pre',
+          fontFamily: 'Fira Code',
```
kyranjamie, line 15: Should fallback to Consolata, monospace etc, most won't have Fira Code installed.

src/components/highlighter/index.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import Prism from 'prismjs';
+import Highlight from 'prism-react-renderer';
+import { Box, Flex, useTheme } from '@blockstack/ui';
+
+import { Token, GetTokenProps, RenderProps } from '@components/highlighter/types';
+import { theme } from '@components/highlighter/prism-theme';
+
+const lineNumberWidth = 60;
+const getLineNumber = (n: number) => (n + 1 < 10 ? '0' : '') + (n + 1);
```
kyranjamie, line 10: This should probs take a total length right, so padding can be 2 0's if it's 100+ long?

src/components/highlighter/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const Tokens = ({
```
kyranjamie, line 12: Can we give `Token` a more explicit name, something like `GrammaticalToken`?  Normally would be fine, but given the context of the app, with NonFungibleTokens and things like this, can get confusing.

src/components/codeblock.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Box } from '@blockstack/ui';
+import css from '@styled-system/css';
+import { Highlighter } from '@components/highlighter';
+
+export const CodeBlock = ({ code }: { code: string }) => {
+  const [editorCode] = useState(code?.toString().trim());
+
+  return (
+    <Box
+      // @ts-ignore
```
kyranjamie, line 11: What's the error without this?

aulneau: css isn't currently a prop that be passed to any element for some reason. It's built into styled-components, and I haven't determined how to type it yet. You should be able to pass `css={css({..})}` and it will turn into a styled-component via the babel plugin.

src/components/highlighter/index.tsx
aulneau: sure, much of this was taken from the way prism and the prism-react-renderer name thing, but your points make sense.

src/components/codeblock.tsx
aulneau: yeah, should add a fallback. We are loading the font, too, so folks will have it.

src/components/codeblock.tsx
kyranjamie: More and more the sx={} seems like a good idea

aulneau: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31245#issuecomment-586480973

aulneau: Yeah, I want to add support for the sx prop, but it's not a straight port over to theme-ui. Involves some more complex adjustments. I'll make an issue to explore it though in blockstack/ui!

PR #89, opened 2020-04-21
feat: perf and a11y updates

aulneau:
#### Before
Screen Shot 2020-04-21 at 3 51 05 PM

#### After
Screen Shot 2020-04-21 at 3 50 25 PM


next.config.js
```diff
@@ -7,25 +11,42 @@ module.exports = {
...
   webpack(config, { dev, isServer }) {
     const splitChunks = config.optimization && config.optimization.splitChunks;
     if (splitChunks) {
       const cacheGroups = splitChunks.cacheGroups;
-      const reactModules = /[\\/]node_modules[\\/](react|react-dom)[\\/]/;
+      const preactModules = /[\\/]node_modules[\\/](preact|preact-render-to-string|preact-context-provider)[\\/]/;
       if (cacheGroups.framework) {
-        cacheGroups.react = Object.assign({}, cacheGroups.framework, {
-          test: reactModules,
+        cacheGroups.preact = Object.assign({}, cacheGroups.framework, {
```
kyranjamie, line 21: Is there a big gain in switching for preact?

server.js
```diff
@@ -0,0 +1,45 @@
...
+  },
+  send: ({ data, res }) => res.send(data),
+});
+
+app.prepare().then(() => {
+  const server = express();
+
+  server.get('/', (req, res) => ssrCache({ req, res, pagePath: '/' }));
+
+  server.get('/txid/:txid', (req, res) => {
```
kyranjamie, line 33: This renders once, and returns cache for the next hour? What about txs that are pending and have a status that'll soon change?

src/common/fonts.ts
```diff
@@ -0,0 +1,15 @@
+import FontFaceObserver from 'fontfaceobserver';
+
+export const handleFontLoading = async () => {
+  const link = document.createElement('link');
+  link.href =
+    'https://fonts.googleapis.com/css2?family=Fira+Code&family=Inter:wght@400;500;600;700&display=swap';
+  link.rel = 'stylesheet';
+  document.head.appendChild(link);
+  const inter = new FontFaceObserver('Inter');
+  const fira = new FontFaceObserver('Fira Code');
```
kyranjamie, line 10: ah, we're loading Fira..

server.js
aulneau: Yeah valid question. We can shorten the time. It drastically improves perf to have a cached version. Seems like right now the perf hit comes from the time it takes to fetch on the server to the sidecar. I wonder what if any kind of caching the sidecar will have?

next.config.js
aulneau: Screen Shot 2020-04-22 at 9 01 47 AM

Lighthouse with React:
Screen Shot 2020-04-22 at 8 58 24 AM

Lighthouse with Preact:
Screen Shot 2020-04-22 at 9 00 00 AM

server.js
kyranjamie: I think it might be best to not use it for tx page, on the hypothesis the average number of page views per tx will be < 1.

PR #95, opened 2020-04-24
Feat/contract call tx page

aulneau:


src/components/token-transfer.tsx
```diff
@@ -137,6 +138,20 @@ const ItemIcon = ({ type }: { type: EventSingle['event_type'] }) => {
       );
   }
 };
+
+const ItemThing = ({ data }: { data: any }) => {
+  if (data.contract_log) {
+    let value = data.contract_log.value;
+    const thing = deserializeCV(Buffer.from(data.contract_log.value.replace('0x', ''), 'hex'));
+
+    if (thing.type === 5 && 'address' in thing) {
```
kyranjamie, line 147: can you make a helper fn that describes what this does?

src/store/transactions/selectors.ts
```diff
@@ -19,3 +20,15 @@ export const selectTransactionLastTxId = createSelector(
   selectTransactionsSlice,
   state => state.lastTxId
 );
+
+export const selectOriginContractSource = (contractName: string) =>
+  createSelector(selectTransactionsSlice, state => {
+    const address = contractName.split('.')[0];
+    const txs = Object.values(state.entities) as Transaction[];
```
kyranjamie, line 27: Would be nice to use `.values` but we can avoid the typecasting doing this, think it's worthwhile as its more type-safe.

```suggestion
    const txs = Object.keys(state.entities).map(tx_id => state.entities[tx_id]);
```

src/store/transactions/selectors.ts
```diff
@@ -19,3 +20,15 @@ export const selectTransactionLastTxId = createSelector(
...
+
+    const contract = txs.find(
+      (tx: Transaction) => tx.tx_type === 'smart_contract' && tx.sender_address === address
+    ) as SmartContractTransaction;
```
kyranjamie, line 31: Should now be `SmartContractTransaction` implictly.

```suggestion
    const contract = txs.find(
      (tx: Transaction) => tx.tx_type === 'smart_contract' && tx.sender_address === address
    );
```

src/components/token-transfer.tsx
aulneau: Yeah def should have renamed this lol

PR #98, opened 2020-04-27
Feat/darkmode

aulneau:


src/pages/_app.tsx
```diff
@@ -8,8 +8,60 @@ import { RootState, initStore } from '@store';
 import { handleFontLoading } from '@common/fonts';
 
 const GlobalStyles = createGlobalStyle`
+@media (prefers-color-scheme: dark) {
+  html, body {
+    --colors-bg: ${
+      // @ts-ignore
+      theme.colors.ink
```
kyranjamie, line 15:
```suggestion
      (theme as any).colors.ink
```

You can typecast theme to `any`, or better, a typing of `theme`. Doing this, I've never come across a need to use `// @ts-ignore`, hope this will fix the error

aulneau: fixed in my other pr :)

PR #103, opened 2020-05-06
feat: add dockerfile

kyranjamie:
`docker build -t blockstack/explorer .`

`docker run -p 3000:3000 blockstack/explorer`

I'm no Docker expert, so may well be missing a trick or two. Seeing as it's just the one service, guess we don't need a compose file?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:alpine
+
+COPY package*.json ./
+RUN yarn
+
+COPY . .
+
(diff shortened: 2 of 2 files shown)


aulneau:
I think generally it's good to try to use the alpine versions of the node docker images for size and such, here is one example I found that seems good -> https://github.com/selfup/nextjs-docker/blob/master/Dockerfile and the one we use for app.co https://github.com/blockstack/app.co/blob/master/Dockerfile

relevant lines (i think)
```
FROM node:alpine
RUN apk add --no-cache libc6-compat
```

zone117x:
Just saw this PR -- FYI if you're interested in minimizing image size, you should run `npm prune --production` after build to remove dev-dependencies.

PR #108, opened 2020-05-09
Feat: better token transfer component

aulneau:
### what this does
This implements a better design for the token transfer events that can be found within a contract.

token-transfer

resolves: https://github.com/blockstack/explorer/issues/87, https://github.com/blockstack/explorer/issues/77


zone117x:
Nice job @aulneau this looks great

PR #115, opened 2020-05-12
feat: contract-call tx page and other improvements

aulneau:
### What this does
This improves the contract call tx page:
- renders out function summary data
- adds post condition component to this page

Screen Shot 2020-05-11 at 3 36 07 PM


aulneau:
@kyranjamie can you take a look at this today?

PR #127, opened 2020-05-15
feat: adding timestamp tooltip to blocktime row

### LINKED ISSUE #120: Block time does not show an exact time stamp
Timestamp does not show an exact time stamp like 11:09:15 it shows the time from  3 hours ago. Should we be showing exact time.

Screen Shot 2020-05-13 at 3 10 00 PM

agraebe:
To fix #120

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "blockstack": "^21.0.0",
     "c32check": "^1.0.1",
     "cacheable-response": "^2.0.8",
-    "dayjs": "^1.8.26",
+    "dayjs": "^1.8.27",
     "express": "^4.17.1",
     "fontfaceobserver": "^2.1.0",
     "formik": "^2.1.4",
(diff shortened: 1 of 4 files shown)


kyranjamie:
I'd favour `toISOString()` for the tooltip personally, to avoid locale-specific ambiguities. All hail ISO 8601 standard.

Side note, we have both `dayjs` and `timeago.js` packages, only really need the one

aulneau:
Thanks for taking a stab at this! When I had been thinking about this change, there were a few goals I had in mind:

- The timestamp component itself should have the tooltip within it and it should conditionally wrap the timestamp
  - allow a prop to be passed such as `noTooltip` to only render the timestamp without the tooltip (rather than `showTooltip` set to true by default, trying to prevent folks from having to pass a false prop `showTooltip={false}`)
- replace the timago.js implementation with the dayjs relative time implementation that kyan did for the recently viewed component, [here](https://github.com/blockstack/explorer/blob/master/src/components/recently-viewed.tsx#L12) and [here](https://github.com/blockstack/explorer/blob/master/src/components/recently-viewed.tsx#L87)
- use `.toISOString()` as kyran requested.
- we should also add in conditional checks for if the `ts` is passed or not

agraebe:
Thanks @aulneau, I'll work on that 👍

agraebe:
@aulneau done

PR #131, opened 2020-05-15
feat: improved recently viewed component (popover component)

aulneau:
### What this does
This uses the new popover component for the recently viewed transactions in the search bar. This gives us a more a11y friendly experience: tabbing, keyboard usage, etc.

You can:
- tab into the search box
  - tab back and forth through all the items
  - use up and down arrows to navigate to items
  - hit enter/return to select one of the items
- hit esc to remove the dropdown/focus of element(s)
- drop down goes away on the small version of the search bar when an error is present

Kapture 2020-05-16 at 17 17 59

This was actually pretty complex and shows how much goes into making things a11y friendly. This `Popover` component will ultimately live in `@blockstack/ui`, but need to battle test it further before bringing it over.


aulneau:
@kyranjamie could you take a look at this PR today?

src/components/popover/use-popover.tsx
```diff
@@ -0,0 +1,177 @@
+import React, { useState, useEffect, useCallback } from 'react';
+import { useEventListener } from '@blockstack/ui';
+import { useControlledHover } from '@common/hooks/use-controlled-hover';
+import { useFocus } from 'use-events';
+import { UsePopoverProps } from '@components/popover/types';
+
+export const usePopover = ({ length, triggerRef }: UsePopoverProps) => {
+  const timeoutRef = React.useRef<number | undefined>(undefined);
+  const removeFocusRef = React.useRef<number | undefined>(undefined);
```
kyranjamie, line 9: Any reason not to use `null` here?  This tends to be my preference, as it's clear, then, something's explictly been set as `null`, rather than just having not been defined and implictly being `undefined`.

aulneau:
@kyranjamie All mine!

src/components/popover/use-popover.tsx
aulneau: yeah should probs be null, i tend to stick with undefined due to ts generally liking it better, and i must have just opted for that here instead of null. i can go through and ensure all the refs are default to null and fall back to it as the option

PR #134, opened 2020-05-17
feat: testnet selector, env vars

aulneau:
### What this does

This enables the ability to easily switch between the mocknet server and the testnet staging server. This functionality needed to be built for the eventual use case of switching between mainnet and testnet. An added benefit of this is now we are passing the env variables via the server, and not when it's built. The variables are loaded within the server and then passed to the redux store.

### todo
- [x] save a cookie with selection and load it on first load

Screen Shot 2020-05-18 at 8 49 52 AM


aulneau:
@kyranjamie can you 👀 this today?

One more feature I think this might need is we could add a `?network=${selectedNetwork}` in the url that the application can parse to make an on-the-fly switch for when folks share links, or at the very least give more information as to why someone might not see a tx if they are on the testnet/mocknet.

Or maybe this should just be a dev mode only feature! Would love your thoughts.

src/store/ui/reducer.tsx
```diff
@@ -1,18 +1,45 @@
...
+  },
 };
 
 export const uiReducer = createReducer(initialState, builder => {
+  builder.addCase(setNetworks, (state, action) => {
+    state.config.network = {
+      ...state.config.network,
+      ...action.payload,
+    };
+  });
```
kyranjamie, line 39: I see this is part of redux-toolkit (checked docs) I find it super counterintuitive to set the state object like this rather than returning an object.

A wider point here: IMO it's a super poor API decision on their part to facilitate this. A reader of this code has to understand a pretty specific implementation of redux-toolkit to know that this isn't mutating state (even if they're familiar with the redux pattern). And what's the uptake? A few cases where some lines of code are saved? I know using too many spred operators gets verbose, but that verbosity is worth not breaking a pattern as widely understood as redux.

There are perfectly good alternatives too, such as the [lens pattern](https://functional.christmas/2019/6), or using `Immer` explicitly (and not behind redux-toolkits implementation). 

Thoughts? Would really love it if we could avoid mutating state like this.

src/common/hooks/use-random-tx.tsx
```diff
@@ -1,19 +1,25 @@
...
 
 export const useNavigateToRandomTx = () => {
+  const { apiServer } = useSelector((state: RootState) => ({
+    apiServer: selectCurrentNetworkUrl(state),
+  }));
   const { addCriticalToast } = useToast();
   const handleRandomTxClick = useCallback(async () => {
     try {
-      await navgiateToRandomTx();
+      await navgiateToRandomTx(apiServer as string)();
```
kyranjamie, line 15: Why is type casing needed here? Is it sometimes undefined? If possible, setting a default in store so it's always defined would be nicer than casting.

src/store/contracts/actions.ts
```diff
@@ -1,9 +1,12 @@
 import { createAsyncThunk, createAction } from '@reduxjs/toolkit';
 import { Contract } from '@models/contract.interface';
 import { fetchContract as fetchContractFromApi } from '@common/api/contracts';
+import { selectCurrentNetworkUrl } from '@store/ui/selectors';
 
-export const fetchContract = createAsyncThunk<Contract, string>('contracts/fetch', async query =>
-  fetchContractFromApi(query)
+export const fetchContract = createAsyncThunk<Contract, string>(
+  'contracts/fetch',
+  // @ts-ignore
```
kyranjamie, line 8: What's the alternative to using ts-ignore here? Perhaps this is a militant statement, but imo we should _never_ be using this.

src/store/sandbox/actions.ts
```diff
@@ -40,9 +41,11 @@ export const eraseIdentity = createAction('account/identity/erase');
 export const fetchAccount = createAsyncThunk<AccountPayload, string>(
   'account/fetch',
   // @ts-ignore
-  async (principal, { rejectWithValue, dispatch }) => {
+  async (principal, { rejectWithValue, dispatch, getState }) => {
+    // @ts-ignore
```
kyranjamie, line 45: 😱

src/store/sandbox/actions.ts
```diff
@@ -87,8 +90,10 @@ export const fetchAccount = createAsyncThunk<AccountPayload, string>(
 export const requestFaucetFunds = createAsyncThunk<Account, string>(
   'account/faucet',
   // @ts-ignore
-  async (principal, { rejectWithValue }) => {
-    const res = await postToSidecar(`/debug/faucet?address=${principal}`);
+  async (principal, { rejectWithValue, getState }) => {
+    // @ts-ignore
```
kyranjamie, line 94: 🙀

src/store/ui/reducer.tsx
aulneau: I think this is pretty subjective and open to opinion. It's more of a code style thing (one that I am very much in favor of, and so are 16k other folks using immer). 

It's not a true mutation, but I understand the concern. I think we could in a different time talk about the pros and cons of using this toolset in the way we are, but it's very much out of scope for this PR.

src/store/sandbox/actions.ts
aulneau: I think this ts-ignore is needed because the thunk action isn't typed correctly but I haven't had time to determine that yet. This pr is the first instance in the app of needing to use a state value within an action (the api server).

src/store/contracts/actions.ts
aulneau: same issue https://github.com/blockstack/explorer/pull/134/files/878ab1d54796294bb3e53a0cd3923367e0b8345e#r426730594

src/store/sandbox/actions.ts
aulneau: also same issue https://github.com/blockstack/explorer/pull/134/files/878ab1d54796294bb3e53a0cd3923367e0b8345e#r426730594

src/store/contracts/actions.ts
kyranjamie, line 8:
```suggestion
export const fetchContract = createAsyncThunk<Contract, string>(
  'contracts/fetch',
```

src/store/contracts/actions.ts
```diff
@@ -1,9 +1,12 @@
...
+  async (query, { getState }) => fetchContractFromApi(selectCurrentNetworkUrl(getState()))(query)
```
kyranjamie, line 9:
```suggestion
  async (query, { getState }) =>
    fetchContractFromApi(selectCurrentNetworkUrl(getState() as any) as string)(query)
```

This casting is preferable to ts-ignore, since even thought it bypasses this particular type issue, it doesn't forgot all type safty on this line.

src/store/transactions/actions.ts
```diff
@@ -4,29 +4,32 @@ import { Contract } from '@models/contract.interface';
...
 import { queryWith0x } from '@common/utils';
 import { fetchContract } from '@store/contracts';
+import { selectCurrentNetworkUrl } from '@store/ui/selectors';
 
 export const fetchTransaction = createAsyncThunk<Transaction[], string>(
   'transaction/fetch',
-  async (query, { dispatch, rejectWithValue }) => {
+  async (query, { dispatch, rejectWithValue, getState }) => {
+    // @ts-ignore
+    const apiServer = selectCurrentNetworkUrl(getState());
```
kyranjamie, line 13:
```suggestion
    const apiServer = selectCurrentNetworkUrl(getState() as any) as string;
```

This isn't so different, but think it's preferable as at least we're being very explicit about what we're not typing.

src/store/sandbox/actions.ts
kyranjamie: The type error here is helpful. The typing says it returns `Account`, but it only returns two of Account's 4 props:

```
export interface Account {
    balance: BN;
    nonce: number;
}
```

modified above

kyranjamie, line 94:
```suggestion
```

src/store/sandbox/actions.ts
```diff
@@ -87,8 +90,10 @@ export const fetchAccount = createAsyncThunk<AccountPayload, string>(
 export const requestFaucetFunds = createAsyncThunk<Account, string>(
```
kyranjamie, line 90:
```suggestion
export const requestFaucetFunds = createAsyncThunk<
  Pick<Account, 'principal' | 'transactions'>,
  string
```

aulneau:
@kyranjamie as we spoke I will make a new PR with the ts-ignore changes

PR #145, opened 2020-05-19
fix: compiler ignores producing incorrect types

kyranjamie:


src/common/sandbox/index.ts
```diff
@@ -87,13 +87,13 @@ export const doGenerateIdentity = async () => {
...
+export const network = (apiServer: string): Partial<StacksNetwork> => ({
   version: TransactionVersion.Testnet,
   chainId: ChainID.Testnet,
   coreApiUrl: withApiServer(apiServer)(),
-  broadcastApiUrl: withApiServer(apiServer)('/v2/transactions'),
-  transferFeeEstimateApiUrl: withApiServer(apiServer)('/v2/fees/transfer'),
-  balanceApiUrl: withApiServer(apiServer)('/v2/accounts'),
+  broadcastEndpoint: withApiServer(apiServer)('/v2/transactions'),
+  transferFeeEstimateEndpoint: withApiServer(apiServer)('/v2/fees/transfer'),
+  accountEndpoint: withApiServer(apiServer)('/v2/accounts'),
```
aulneau, line 96: These don't line up with: https://github.com/blockstack/stacks-transactions-js/blob/master/src/network.ts#L3

kyranjamie: Didn't get around to finishing this.

The issue here is that if we wanna type something as `StacksNetwork` we need all of those properties to conform to the interface.

kyranjamie: The `Partial` I added doesn't work, because it's being passed to an object that requires `StacksNetwork` (build fail)

aulneau:
Closing this due to age. Some of this has already been implemented and others will be in a future PR.

PR #166, opened 2020-05-27
feat: initial connect integration

aulneau:
### What this does

This updates the sandbox to require blockstack authentication to do anything. Eventually we will be handling all the address related stuff in the authenticator, but for now it's a combo of generating a keypair and logging into the authenticator.


src/common/hooks/use-user-session.tsx
```diff
@@ -1,8 +1,11 @@
 import * as React from 'react';
 import { UserSession, AppConfig } from 'blockstack';
+import { UserData } from 'blockstack/lib/auth/authApp';
 
 export const useUserSession = () => {
-  const [userData, setUserData] = React.useState({ loaded: false });
+  const [userData, setUserData] = React.useState<{ loaded: boolean } & Partial<UserData>>({
```
kyranjamie, line 6: Now this creates a type union, would recommend creating a new type.

src/components/sandbox/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { truncateMiddle, microToStacks } from '@common/utils';
+import { useUserSession } from '@common/hooks/use-user-session';
+import { useDebugState } from '@common/sandbox';
+import { Tabs } from '@components/sandbox/tabs';
+import { Text, Title } from '@components/typography';
+import { PageWrapper } from '@components/page';
+import { Card } from '@components/card';
+import { Meta } from '@components/meta-head';
+
+const SignedOutView = ({ onClick }: any) => {
```
kyranjamie, line 15: Ensures it's typed as a fn that return `any`

```suggestion
const SignedOutView = ({ onClick }: { onClick: () => any }) => {
```

src/common/hooks/use-user-session.tsx
aulneau: This is updated/changed in a different branch, will avoid changing here.

src/components/sandbox/page.tsx
aulneau: This is updated/changed in a different branch, will avoid changing here.

PR #171, opened 2020-05-28
fix: auto split contract name, add post condition mode select

aulneau:
### Contract names
This adds a nice to have for when using the contract call functionality of the sandbox. Typically you'd copy the fully realized name of the contract, eg `ST14Z2G3JTG8805N2CCVXG85XGFJP6PME385X69JG.hello`. You can now paste this into the address field, and it will automatically split the address and name, filling both fields with the values.

Kapture 2020-05-28 at 11 42 49

### Post condition mode
This also adds the drop down to all function calls for selecting the post condition mode. Eventually we will have a post conditions component that will allow the user to add any possible combination of post conditions. 

Screen Shot 2020-05-28 at 11 45 28 AM

### DIFF
--- a/src/components/sandbox/contract-call.tsx
+++ b/src/components/sandbox/contract-call.tsx
@@ -15,14 +15,14 @@ import { useRef } from 'react';
 
 const Functions = ({ abi, contractName, contractAddress }: any) => {
   return abi.functions.map((func: any) => {
-    return (
+    return func.access !== 'private' ? (
       <Function
         contractName={contractName as string}
         contractAddress={contractAddress as string}
         func={func}
         key={func.name}
       />
-    );
+    ) : null;
   });
 };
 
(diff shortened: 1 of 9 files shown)


src/components/sandbox/contract-call.tsx
```diff
@@ -111,35 +125,54 @@ export const ContractCall = (props: any) => {
...
-                        <Box>
-                          <Button ref={buttonRef as any} type="submit" isLoading={isLoading}>
-                            Fetch
-                          </Button>
-                        </Box>
-                      </Stack>
-                    </Flex>
-                  </Box>
-                )}
+                {({ handleSubmit, setValues }) => {
```
kyranjamie, line 128: forkik exposes hooks too, to avoid nesting in the jsx, i saw recently

aulneau: I was reading and using the `useFormik` hook but my understanding was that the render props method is the main way, and they don't recommend using the hook as it's more for internal usage only. I do prefer the hook.

PR #172, opened 2020-05-28
fix: account for new status options

aulneau:
### What this does

This takes into account two new options for the status of a transaction:  `abort_by_response` `abort_by_post_condition`. This PR only makes sure that nothing breaks. In a different PR I will address adding messaging to help folks understand better why the transaction failed. Related: https://github.com/blockstack/stacks-blockchain-sidecar/pull/126

### DIFF
--- a/src/components/status.tsx
+++ b/src/components/status.tsx
@@ -2,18 +2,27 @@ import * as React from 'react';
 import { Flex, Text, Box, Spinner, BoxProps } from '@blockstack/ui';
 import { CheckmarkCircleIcon, ExclamationMarkCircleIcon } from '@components/svg';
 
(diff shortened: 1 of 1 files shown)


src/components/status.tsx
```diff
@@ -2,18 +2,27 @@ import * as React from 'react';
 import { Flex, Text, Box, Spinner, BoxProps } from '@blockstack/ui';
 import { CheckmarkCircleIcon, ExclamationMarkCircleIcon } from '@components/svg';
 
-export type Statuses = 'success' | 'pending' | 'failed';
+export type Statuses =
```
kyranjamie, line 5: This is in the types repo: `TransactionStatus`

PR #179, opened 2020-06-02
fix: popover wrapper ref

aulneau:


aulneau:
@kyranjamie sure, previously the wrong ref was used for the onClickOutside hook used here. Any click outside of the trigger would cause the dropdown to be removed, preventing the click of items in the drop down. Now anything clicked within the wrapper itself (holding the list too) will not cause it to close. https://github.com/blockstack/explorer/pull/179/files#diff-034bc34d2aaba7b645a7937c308cb337R188

PR #217, opened 2020-07-22
build: update connect

hstove:
Updates connect to latest, and updates the `authOptions` to use `onFinish` and the default `authOrigin`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "version": "1.0.0",
   "dependencies": {
-    "@blockstack/connect": "^2.8.0",
+    "@blockstack/connect": "^2.8.6",
     "@blockstack/rpc-client": "^0.2.0-alpha.0",
     "@blockstack/stacks-transactions": "^0.4.4",
     "@blockstack/ui": "^2.8.0",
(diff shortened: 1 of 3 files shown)


aulneau:
@hstove I just merged a PR, can you rebase off `master` with these changes before I review?

hstove:
@aulneau yep, all set.

aulneau:
@hstove I tried the preview link and there is something off with Connect:
Screen Shot 2020-07-22 at 7 36 19 PM

hstove:
@aulneau cool, looks like the resolution fixed it. Weird issue.

Sometimes the way node/npm supports multiple versions of a dep is nice, and sometimes it can be a pain in the butt (like here). I remember "fun" times in Ruby trying to coerce my gems so that they'd be able to resolve to the same version of a dependency.

aulneau:
@hstove there is still a bug here that once someone logs in, the connect modal does not dismiss. Could it be a regression in connect somewhere?

markmhx:
@aulneau mind creating an issue in the UX repo with the problem you see and linking to here?

agraebe:
updating this issue based on the sprint meeting: this should be based on the latest version (with stenciljs)

PR #301, opened 2021-01-12
fix: breaking connect change around stx address in profile

hstove:
As some requested, the `profile` now includes `stxAddress: { testnet, mainnet }`. Unfortunately this broke the explorer.


aulneau:
I've got a local version with this fix, closing

PR #308, opened 2021-01-13
fix: update connect

hstove:
This includes the new logic, where the user has to have the extension installed.

Also in the future if you need to check for the extension, we have `import { isStacksWalletInstalled } from '@stacks/connect'`


aulneau:
@hstove can you rebase?

PR #333, opened 2021-01-16
Improved transactions list + more

aulneau:
changelog-001

[Click here for staging](https://explorer-git-feat-improved-transactions.blockstack.now.sh/)

### Updates to the transactions list component

This update brings some improvements to the component that displays lists of transactions on the homepage and on the
transactions page. Previously you were only able to see a limited number of transactions from the mempool in either
view. Now you are able to see all items currently pending, paginate through them, and filter between types of
transactions. This update also changes the way we are filtering transactions: previously there was a naive
implementation that simply filtered through the data returned. This meant that if the limit was set to 50, and all 50 of
the returned items were of one type, filtering to view others would remove all transactions from view. Now the data is
re-fetched with the selected types. This gives a much better experience, allowing users to see all available transactions
of a given set of types.

### Function summaries on pending Contract Call transactions

We had previously been reliant on some data that was only available after a transaction was confirmed. The component
that displays the function summary of a contract-call transaction has been refactored to use only the data that is
available before and after confirmation. This is useful when doing many calls to the same function with different
arguments passed -- previously it would have been hard to keep track of which contract-call transaction had which
arguments.

### Other fixes

- When viewing a `stack-stx` function call transaction, the `pox-addr` parameter is now parsed and rendered as the BTC
  address it represents.
- Fixed a bug where the transaction list component would display "Load more transactions" when in-fact there were no more transaction to load.
- Added Fathom tracking
- Improvements + fixes for the sandbox and logging in with the Stacks Wallet Extension
- Fixes bug where stale data would persist in the transactions list of an address page when navigating to another address page

## For reviewers

This PR starts the refactor to using `react-query` for data fetching in place of `useSWR`. The main rationale behind this is the way that `react-query` handles SSR data fetching. They use a de/hydration pattern where you can fetch data on the server, dehydrate it, and the client will hydrate the data based off of keys. [You can read more about that here.](https://react-query.tanstack.com/guides/ssr#using-hydration) All in all they are very similar, but I generally feel as though I prefer react-query after using it a bit. 

I have not done all the work to remove the `useSWR` hooks -- I wanted to get the features I was working on out of the door before spending too much time refactoring working code. I should have a PR up in the next few days refactoring away the other hooks.

To test out filter functionality: navigate to `/transactions`, hit the 'Confirmed transactions' tab and use the filter panel to the right of the section header.


src/common/hooks/use-chain-mode.ts
```diff
@@ -4,41 +4,43 @@ import { useModal } from '@common/hooks/use-modal';
...
 
 // kind of like useState, but for the query param state
 export const useChainMode = (): [ChainMode, SetChainMode] => {
   const router = useRouter();
+  const { setAuthResponse } = useAuthState();
 
-  const authResponse = router.query?.authResponse
-    ? { authResponse: router.query?.authResponse }
-    : {};
+  const setChainMode = async (chain: ChainMode) => {
```
kyranjamie, line 17: How come this is no longer in a usecallback?

src/common/hooks/use-set-network-mode.ts
```diff
@@ -0,0 +1,9 @@
+import { NetworkModes } from '@common/types/network';
+import { useMemo } from 'react';
+import { setCurrentNetworkMode } from '@common/app-helpers';
+
+export const useSetNetworkMode = (networkMode: NetworkModes) => {
+  useMemo(() => {
+    setCurrentNetworkMode(networkMode);
+  }, []);
```
kyranjamie, line 8: Shouldn't pass in networkMode? 🤔

src/common/lib/transactions.ts
```diff
@@ -0,0 +1,53 @@
...
+import { generateTransactionsQueryPath } from '@common/lib/utils';
+import type { FetchTransactionsBase } from '@common/lib/types';
+
+/**
+ * Fetch list of confirmed or pending txs
+ */
+export function fetchTransactionsList(apiServer: string) {
+  return async function (
+    path: string
+  ): Promise<TransactionListResponse | MempoolTransactionListResponse | any> {
```
kyranjamie, line 15: Doesn't passing `| any` render any additional type information worthless? 🤔

src/components/app-container.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { AppWrapper } from '@components/app-init';
+import { useChainModeEffect } from '@common/hooks/use-chain-mode';
+
+interface AppContainerProps {
+  isHome?: boolean;
+}
+
+export const AppContainer: React.FC<AppContainerProps> = props => {
+  const { children, isHome } = props;
```
kyranjamie, line 10: Love destructuring props on a new line 😂

kyranjamie (approved):

src/common/hooks/use-set-network-mode.ts
aulneau: ah, no, we don't need to in this case. Normally we would but in this case we want it to run only one time. It is apart of this trick to set a recoil value on initial load from a value fetched on the server https://github.com/blockstack/explorer/blob/feat/improved-transactions/src/common/app-helpers.ts#L13

src/common/hooks/use-chain-mode.ts
aulneau: It was causing some issues -- might have removed it while debugging

aulneau: I have a new pr that improves some of this, will make this change in that PR

PR #359, opened 2021-02-03
fix: update connect to canary version

hstove:
Update `@stacks/connect` version for Canary


aulneau:
This is taken care of in my sandbox pr, closing

PR #395, opened 2021-02-26
fix: use correct title in meta

friedger:
This PR
* changes the title in Recent Blocks to exactly that


aulneau:
Awesome, thank you!

PR #399, opened 2021-03-01
fix: fathom

aulneau:
This fixes fathom to actually work. Something was up before where it wasn't registering correctly.

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -12,7 +12,6 @@ const {
   NEXT_PUBLIC_CONNECT_AUTH_ORIGIN,
   NEXT_PUBLIC_TESTNET_API_SERVER,
   NEXT_PUBLIC_MAINNET_API_SERVER,
-  NEXT_PUBLIC_FATHOM_ID,
 } = publicRuntimeConfig;
 
 const getNumber = (query?: string): number | undefined =>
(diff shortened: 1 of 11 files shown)


src/common/constants.ts
```diff
@@ -92,7 +92,7 @@ export const DEFAULT_NETWORK_LIST = [
   },
 ];
 
-export const FATHOM_ID = NEXT_PUBLIC_FATHOM_ID || process.env.NEXT_PUBLIC_FATHOM_ID;
+export const FATHOM_ID = 'YAKFSIOZ' || NEXT_PUBLIC_FATHOM_ID || process.env.NEXT_PUBLIC_FATHOM_ID;
```
kyranjamie, line 95: Do we want to be hardcoding this? None of the rest of this expression will ever be reached.

src/common/hooks/use-fathom.ts
```diff
@@ -1,24 +1,27 @@
...
 
 export const useFathom = () => {
   const router = useRouter();
 
   useEffect(() => {
-    Fathom.load(FATHOM_ID, {
-      includedDomains: ['https://explorer.stacks.co'],
-    });
+    if (process.env.NODE_ENV !== 'development' && typeof document !== 'undefined') {
+      Fathom.load('YAKFSIOZ', {
```
kyranjamie, line 14: duplicated

kyranjamie (approved):

Issue #322, opened 2021-01-15
Account and transaction page: Show nonce


agraebe:
Developers might want to confirm the nonce for an account, either before or after transactions broadcasting. It would be great if the explorer could show the nonce.

We should review how accurate we could be when displaying the nonce (with multiple txs in the same block and macroblocks, this might make it not very reliable).

friedger:
Looks like fixed

PR #420, opened 2021-03-28
feat: add nonce to tx and account page

agraebe:
fixing #322

## Description

1. Motivation for change
Oftentimes, transactions are pending for a long time due to nonce issues. It is not possible to review the nonces in the explorer, so the API is used.

2. What was changed
Adding a nonce field to the account and transaction page

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Are documentation updates required?
No

## Testing information
Tested locally and compared to API results.

## Checklist
- [x] Code is commented where needed
- [ ] Unit test coverage for new or modified code paths
- [ ] `npm run test` passes
- [ ] Changelog is updated
- [x] Tag @aulneau

### DIFF
--- a/src/components/transaction-details.tsx
+++ b/src/components/transaction-details.tsx
@@ -58,9 +58,7 @@ const AddressComponent = React.memo(({ principal }: { principal: string }) => {
 });
 
 const NoncesComponent = React.memo(({ nonce }: { nonce: number }) => {
-  return (
-    <Text display="block">{nonce}</Text>
-  );
+  return <Text display="block">{nonce}</Text>;
 });
 
 const getSenderName = (txType: Transaction['tx_type']) => {
--- a/src/components/transaction-details.tsx
+++ b/src/components/transaction-details.tsx
@@ -57,6 +57,14 @@ const AddressComponent = React.memo(({ principal }: { principal: string }) => {
   );
 });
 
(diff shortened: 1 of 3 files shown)


src/pages/address/[principal].tsx
```diff
@@ -114,6 +121,7 @@ const SummaryCard = ({ principal, hasTokenBalances, data }: any) => {
 
 interface AddressPageData {
   principal: string;
+  info: AccountDataResponse;
```
Comment (aulneau) [inline src/pages/address/[principal].tsx:124]: this could be more descriptive, eg `accountInfo`, or if you are only using the nonce, return just that as `nonce`

src/components/transaction-details.tsx
```diff
@@ -57,6 +57,14 @@ const AddressComponent = React.memo(({ principal }: { principal: string }) => {
   );
 });
 
+const NoncesComponent = React.memo(({ nonce }: { nonce: number }) => {
+  return (
+    <Box>
+      <Text>{nonce}</Text>
+    </Box>
```
aulneau, line 64:
```suggestion
    <Text display="block">{nonce}</Text>
```

aulneau (approved):

agraebe:
@aulneau it looks like this is not yet in production - is there a manual deployment we should do?

PR #423, opened 2021-03-29
feat: garbage collection note

### LINKED ISSUE #422: feat: show garbage collection info for pending transactions
**Is your feature request related to a problem? Please describe.**
Mempool transactions get garbage-collected after 256 blocks. It would be great if the explorer could show that a mempool transactions (pending for an extended period) will likely be removed from the mempool in X blocks.

**Describe the solution you'd like**
A note on the transaction page for a pending transaction. Maybe only show the note after 128 blocks (~1 day pending, ~1 day to collection). Going a step further, an explanation for pending txs could be provided: https://github.com/blockstack/explorer/issues/340

**Describe alternatives you've considered**
- [x] update documentation: https://github.com/blockstack/docs/pull/1124

**Additional context**
Integration partners would be confused about pending transactions. Most of them would be due to nonce issues. Nevertheless, the garbage collection behavior was surprising.

agraebe:
Fixes #422

### DIFF
--- a/src/components/page.tsx
+++ b/src/components/page.tsx
@@ -16,6 +16,7 @@ import {
   SITE_NOTICE_BANNER_MESSAGE,
   SITE_NOTICE_ENABLED,
 } from '@common/constants';
+import { useNetworkMode } from '@common/hooks/use-network-mode';
 
 type PageProps = {
   notice?: { label?: string; message?: string };
@@ -32,7 +33,10 @@ type PageWrapperProps = {
 export const PageTop: React.FC<TitleProps> = ({ tx, ...props }) => {
   const status = tx.tx_status;
   const failed = status === 'abort_by_response' || status === 'abort_by_post_condition';
-  const longPending = dayjs().diff(dayjs.unix(tx.receipt_time), 'h') > 24;
+  const networkMode = useNetworkMode();
+  // for testnet, show after 4 hours. for mainnet, show after 24 hours
+  const longPending =
+    dayjs().diff(dayjs.unix(tx.receipt_time), 'h') > (networkMode === 'testnet' ? 4 : 24);
 
   const failedMessage =
     status === 'abort_by_response'
--- a/src/components/page.tsx
+++ b/src/components/page.tsx
@@ -44,7 +44,7 @@ export const PageTop: React.FC<TitleProps> = ({ tx, ...props }) => {
       : 'This transaction would have succeeded, but was rolled back by a supplied post-condition.';
 
   const longPendingMessage =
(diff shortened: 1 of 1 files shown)


src/components/page.tsx
```diff
@@ -31,12 +33,19 @@ type PageWrapperProps = {
...
+  const longPending =
+    dayjs().diff(dayjs.unix(tx.receipt_time), 'h') > (networkMode === 'testnet' ? 4 : 24);
 
   const failedMessage =
     status === 'abort_by_response'
       ? 'This transaction did not succeed because the transaction was aborted during its execution.'
       : 'This transaction would have succeeded, but was rolled back by a supplied post-condition.';
 
+  const longPendingMessage =
+    'Transactions that cannot be confirmed within 256 blocks will eventually be garbage-collected.';
```
kyranjamie, line 47:
```suggestion
    'Transactions that cannot be confirmed within 256 blocks are eventually garbage-collected.';
```

kyranjamie: Is garbage collection a term we're using for stale unconfirmed txs?

agraebe: good point. maybe there's a better way to communicate that. I'm assuming some end-users might look at this and not be familiar with garbage collection. We could rather say "... are eventually removed from the mempool."

cc @markmhx or @aulneau ?

agraebe: Combining the feedback Kyran had earlier, here's what I changed it to:

```
'Transactions that cannot be confirmed within 256 blocks are eventually canceled automatically.'
```

(adding "eventually" because it is not exactly known when the collection will happen)

markmhx (approved):

agraebe:
@kyranjamie would love a quick review 🙏🏼

src/components/page.tsx
```diff
@@ -31,12 +33,19 @@ type PageWrapperProps = {
 export const PageTop: React.FC<TitleProps> = ({ tx, ...props }) => {
   const status = tx.tx_status;
   const failed = status === 'abort_by_response' || status === 'abort_by_post_condition';
+  const networkMode = useNetworkMode();
+  // for testnet, show after 4 hours. for mainnet, show after 24 hours
+  const longPending =
+    dayjs().diff(dayjs.unix(tx.receipt_time), 'h') > (networkMode === 'testnet' ? 4 : 24);
```
kyranjamie, line 39: Would be nice to use variables for the 4 and 24 values with descriptive names

kyranjamie (approved):

src/components/page.tsx
agraebe: done!

Issue #418, opened 2021-03-26
Enable select and copy of balance values onto one line


markmhx:
This is a bit of a nit, but I've found myself a few times now selecting the balance on an address page and copying it to my clipboard only to see it pasted on several lines, apparently since the various parts of the balance are rendered in separate HTML tags.

It'd be handy to have the entire selection copied to one line so they're easier to paste elsewhere.

![image](https://user-images.githubusercontent.com/28991/112649155-76653380-8e4a-11eb-8f10-980c00f9c6a7.png)

![image](https://user-images.githubusercontent.com/28991/112649199-80873200-8e4a-11eb-89f9-8acb31350fa1.png)

andresgalante:
This is an HTML issue and really easy to fix

andresgalante:
we'll fix this on the new design

PR #437, opened 2021-05-04
fix: select the balance with one click

OscarVelandia:
## Description

As an interested in the project, I would like to improve the text selection in the address balance. This is needed because it isn't possible to select the balance with click or double click.

This PR was done taking into account the `#418`,  I tested it in windows 10 with notepad, VSCode and Sticky Notes, but I can't reproduce the multi-line paste issue, so, I did an improvement in the select

For details refer to issue #418 by @markmhx

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

## Does this introduce a breaking change?
No

## Are documentation updates required?
No

## Testing information

1. Go to the page http://localhost:3000/address/ST2X2FYCY01Y7YR2TGC2Y6661NFF3SMH0NGXPWTV5?chain=testnet
2. select the STX Balance.

Expected result: it should select all the text.

### DIFF
--- a/src/components/balances/stx-balance-card.tsx
+++ b/src/components/balances/stx-balance-card.tsx
@@ -126,7 +126,6 @@ export const StxBalances: React.FC<StxBalancesProps> = ({
               <Circle bg={color('brand')} mr="base">
                 <StxInline color="white" size="22px" />
               </Circle>
-              {console.log(totalBalance)}
               <Stack spacing="tight" pr="base">
                 <BalanceItem fontWeight="500" color={color('text-title')} balance={totalBalance} />
                 <Caption>Total balance</Caption>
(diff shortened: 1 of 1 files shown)


src/components/balances/stx-balance-card.tsx
```diff
@@ -126,6 +126,7 @@ export const StxBalances: React.FC<StxBalancesProps> = ({
               <Circle bg={color('brand')} mr="base">
                 <StxInline color="white" size="22px" />
               </Circle>
+              {console.log(totalBalance)}
```
aulneau, line 129: this should be removed

aulneau (changes requested):
Nice, thanks for this! just one comment and then we're good

src/components/balances/stx-balance-card.tsx
OscarVelandia: Done

PR #440, opened 2021-05-17
chore: add hiro headers

aulneau:
resolves https://github.com/blockstack/explorer/issues/413


src/common/constants.ts
```diff
@@ -62,6 +62,8 @@ export const DEFAULT_MAINNET_SERVER =
   process.env.NEXT_PUBLIC_MAINNET_API_SERVER ||
   'https://stacks-node-api.stacks.co';
 
+export const VERSION = config?.VERSION || process.env.VERSION;
```
kyranjamie, line 65: Isn't version already set as `process.env.VERSION` already in the next config?

kyranjamie (approved):

src/common/constants.ts
aulneau: This is a requirement for the variable to exists in both the SSR and client side contexts

PR #469, opened 2021-06-24
Cleanup data fetching and state

aulneau:
## What this does

Previously we had been using a combination of three different tools for data fetching / state of external data, [swr](https://swr.vercel.app/), [react-query](https://react-query.tanstack.com/), and [recoil](https://recoiljs.org/). This PR refactors how we are fetching data and how we are using that data in state, moving from recoil as the state management layer to [jotai](https://jotai.pmnd.rs/), like we have in the stacks web wallet.

`react-query` is amazing, and there is a really tight integration of it into jotai, and that is what we're using for all data and state from this PR.

### Notable changes
- removes recoil and swr from the application
- normalizes how we are fetching data throughout the application
- adds jotai and uses react-query for all data fetching
- uses all the `@stacks/stacks-blockchain-api-client` clients (the explorer was made before this was a thing)
- updates the token unlocking feature to use data from our primary api, where previously it was using the legacy explorer API
- refactors and cleans up the search functionality
- significant performance improvements
- lots of dead code removal

## Data fetching on pages
Because this is a next.js application, we need to funnel all of our data fetching through pages. Previously we were using a combination of the suggested method for SSR data handling in react query, which basically consists of fetching the queries on the server and then passing a dehydrated (serialized) response to the query provider component. In addition to that, we were doing other fetches and sending that data to useSWR hooks, and even setting some recoil atoms on mount (which is not really ideal).

Now we are doing something similar, but we are fetching our queries on the server and using the `Provider` that is exported from `jotai` to set an initial data value for a given query. For each page, we have a set of queries for the data we'll need fetched, and you can view all of them in this new folder [`page-queries`](https://github.com/blockstack/explorer/tree/feat/cleanup-data/src/common/page-queries). 

An important point: we're making heavy use of [the `query-key` pattern](https://react-query.tanstack.com/guides/query-keys) that react-query uses, so every query we have has a key that is related to it, and helper methods to construct the correct key. I'd recommend reading [this post about how powerful query keys are.](https://tkdodo.eu/blog/effective-react-query-keys)

Let's take the home page as an example of how our data fetching works. Here is the [homepage component](https://github.com/blockstack/explorer/blob/feat/cleanup-data/src/pages/index.tsx#L36). You can view [the queries for the home page here.](https://github.com/blockstack/explorer/blob/feat/cleanup-data/src/common/page-queries/home.ts)

Our home page component is pretty straightforward:
```tsx
import React from 'react';
import { Grid } from '@stacks/ui';
import { Meta } from '@components/meta-head';
import { HomePageTop } from '@components/home-page-top';
import { TabbedTransactionList } from '@components/tabbed-transaction-list';
import { BlocksList } from '@features/blocks-list';

import { DEFAULT_LIST_LIMIT_SMALL } from '@common/constants';
import { withInitialQueries } from '@common/with-initial-queries';
import { getHomePageQueries } from '@common/page-queries/home';

import type { NextPage } from 'next';

const Home: NextPage = () => {
  return (
    <>
      <Meta />
      <HomePageTop />
      <Grid
        mt="extra-loose"
        gap="extra-loose"
        gridTemplateColumns={['100%', '100%', 'calc(60% - 32px) 40%']}
        width="100%"
      >
        <TabbedTransactionList limit={DEFAULT_LIST_LIMIT_SMALL} />
        <BlocksList limit={DEFAULT_LIST_LIMIT_SMALL} />
      </Grid>
    </>
  );
};

Home.getInitialProps = () => {
  return { isHome: true };
};

export default withInitialQueries(Home)(getHomePageQueries);
```

It's a standard next.js page with some components, and it's got a `getInitialProps` function to return the prop `isHome`. All of our magic is happening with the higher order function `withInitialQueries` which takes a NextPage as the first argument, and returns a function that takes a few other arguments, the first being `getQueries`, which you can see below:

```ts
import { GetQueries } from 'jotai-query-toolkit/nextjs';
import { getApiClients } from '@common/api/client';
import { getTxQueryKey, TransactionsListResponse } from '@store/transactions';
import { DEFAULT_LIST_LIMIT_SMALL } from '@common/constants';
import { BlocksListResponse, getBlocksQueryKey } from '@store/blocks';

export const getHomePageQueries: GetQueries = async ctx => {
  const { transactionsApi, blocksApi } = await getApiClients(ctx);
  return [
    [
      getTxQueryKey.confirmed(DEFAULT_LIST_LIMIT_SMALL),
      async () => {
        return (await transactionsApi.getTransactionList({
          limit: DEFAULT_LIST_LIMIT_SMALL,
          offset: 0,
        })) as TransactionsListResponse;
      },
    ],
    [
      getTxQueryKey.mempool(DEFAULT_LIST_LIMIT_SMALL),
      async () => {
        return (await transactionsApi.getMempoolTransactionList({
          limit: DEFAULT_LIST_LIMIT_SMALL,
          offset: 0,
        })) as TransactionsListResponse;
      },
    ],
    [
      getBlocksQueryKey.confirmed(DEFAULT_LIST_LIMIT_SMALL),
      async () => {
        return (await blocksApi.getBlockList({
          limit: DEFAULT_LIST_LIMIT_SMALL,
          offset: 0,
        })) as BlocksListResponse;
      },
    ],
  ];
};
```

At the most basic level, this is a function that returns an array of tuples, where the first item in the tuple is a query key, and the second is a fetcher method which will return the data associated with that query key.

The `GetQueries` type is really great because it gives us a lot of options around how and what we use to fetch our data. the first argument that gets passed to a `getQueries` function is the `NextPageContext` object, which allows us to use things from the server to inform our queries, such as params or cookies.

For the homepage, we're using the context in our `getApiClients` method, which will extract the current network a user has set (if any) and use that as the baseUrl for all our API clients.

The higher order function, `withInitialQueries ` takes the results of these queries and automatically feeds them to a `<Provider>` component that wraps the page the function is used on. This makes it so we can use any of these queries (so long as they are used in a component that is a child of the page) and have the data on first load.

In addition to that, `withInitialQueries ` will automatically check the cache of queries from `react-query`, and if the key has data associated with it, it will use that in place of doing an additional fetch. This saves us so much time and really improves the performance of the application.

## Devtools integrations
With the use of react-query and jotai, we're able to tap into some more advanced dev tools to help debug the current state of the application at any given time. react-query has a very robust devtool they offer which is enabled by default in dev mode. Jotai is able to connect atoms to the redux-devtools extension, so you can inspect the state of the atoms.

Screen Shot 2021-07-08 at 1 48 31 PM

Screen Shot 2021-07-08 at 1 49 11 PM


src/common/hooks/search/use-recent-items.tsx
```diff
@@ -1,11 +1,13 @@
 import * as React from 'react';
-import { useRecoilState } from 'recoil';
-import { searchRecentlyViewedItemsState } from '@store/search';
+
+import { searchRecentlyViewedItemsState } from '@store/recoil/search';
```
fbwoolf, line 3: What does `recoil` mean here?

src/common/api/client.ts
```diff
@@ -0,0 +1,89 @@
...
+    const url = new URL(context.url);
+    url.searchParams.set('unanchored', 'true');
+    return Promise.resolve({
+      init: context.init,
+      url: url.toString(),
+    });
+  },
+};
+
+// we use to to create our api client config on both the server and client
```
fbwoolf, line 65: Just marking a duplicate word in this comment, `to to`.

src/common/hooks/search/use-recent-items.tsx
aulneau: this folder should be moved up, I'll do that after your initial review -- it was part of my refactor to move all recoil state to its own folder, but now it's using jotai

src/common/page-queries/block-hash.ts
```diff
@@ -0,0 +1,61 @@
...
+// this will get us our query props data from the react-query cache if it exists
+// to prevent duplicate fetching when we don't always need to :)
+function getBlockPageCachedQueryProps(ctx: NextPageContext, queryClient: QueryClient) {
+  const blockHash = getBlockHashFromCtx(ctx);
+  const queryKey = getBlocksQueryKey.single(blockHash);
+  return getSingleCachedQueryData<Block>(queryKey, queryClient);
+}
+
+// this is our function for fetching the transaction being requested
+// the transaction (if found) will be fed as context/props to the getQuerys function
```
fbwoolf, line 26: I think this is supposed to be `block` in comments not `transaction`.

src/components/app-config.tsx
```diff
@@ -1,50 +1,21 @@
...
-                {children}
-              </AppContainer>
-            </CacheProvider>
-          </Hydrate>
-        </QueryClientProvider>
-      </RecoilRoot>
-    </SWRConfig>
-  )
+export const AppConfig: React.FC<AppConfigProps> = ({ children, isHome, fullWidth }) => (
+  <RecoilRoot>
```
fbwoolf, line 14: I see a bit of recoil is being left, I thought we were totally removing Recoil?

src/components/stacking.tsx
```diff
@@ -1,88 +1,88 @@
 import * as React from 'react';
 import { Box, Flex, color, Stack, StxInline, Circle } from '@stacks/ui';
 import { Caption, Text, Link, Title } from '@components/typography';
-import { useStacksInfo } from '@common/hooks/use-stacks-info';
+// import { useStacksInfo } from '@common/hooks/use-stacks-info';
```
fbwoolf, line 4: Did you mean to leave this file commented out?

src/components/app-config.tsx
aulneau: I did not remove recoil from the sandbox -- that is a bit bigger and this PR is already larger than I hoped for. I'll work on that in a follow up PR

src/components/stacking.tsx
aulneau: ah thank you! i will finish that up

src/features/account-transaction-list.tsx
```diff
@@ -0,0 +1,24 @@
+import { Box } from '@stacks/ui';
+import { InfiniteTransactionsList } from '@components/infinite-item-list';
+import { Section } from '@components/section';
+import * as React from 'react';
+import { useAccountInViewTransactions } from '../hooks/currently-in-view-hooks';
```
fbwoolf, line 5: Should we make an `@hooks` alias?

aulneau: yep! I started moving hooks out of `@common/hooks` to `./src/hooks`, but wanted to wait until we get our two PRs merged before doing a larger move like that to avoid conflicts

fbwoolf (approved):
Looks really great, @aulneau! Lots of work. I just tried to note some minor cleanup, but nothing much jumped out at me. Will be interesting to rebase on this! 😄 One thing, I've been testing the deployed version and I can't seem to switch off mainnet in the Network dropdown but the url shows `=testnet`.

aulneau:
> Looks really great, @aulneau! Lots of work. I just tried to note some minor cleanup, but nothing much jumped out at me. Will be interesting to rebase on this! 😄 One thing, I've been testing the deployed version and I can't seem to switch off mainnet in the Network dropdown but the url shows `=testnet`.

Awesome, thanks for the review! yeah I think I just have a couple things left
- error pages / states
- network switching
- stacking card on address page

src/components/app-config.tsx
fbwoolf: That makes sense, got it, thx!

kyranjamie (approved):
Thanks for walk through of the changes.

Huge PR so I kinda skim-reviewed, but LGTM. Let me know you want me to do a more thorough check.

markmhx:
@aulneau does this need QA or can move directly to production deployment?

aulneau:
> @aulneau does this need QA or can move directly to production deployment?

yep -- this should do a full round of QA. I also have a couple tasks to wrap up today:
- test to ensure sandbox is working as expected (didn't change anything)
- error pages

Eshwari007:
@aulneau I am assuming with the clean up data fetching and state which has been implemented in Explorer,a through end to end regression will be apt from QA end.Do correct me if you want to me test anything in specific?

markmhx:
@aulneau @Eshwari007 It's not clear to me that this PR is ready for QA since it's still in "In development" on the board and @aulneau mentioned above needing to finish a couple of tasks?

Eshwari007:
@markmhx Agree .Prior to testing explorer for microblocks feature integration,team decided that I should test this PR hence @aulneau gave me the green signal to test them yesterday via our userx- tech channel.

aulneau:
> @markmhx Agree .Prior to testing explorer for microblocks feature integration,team decided that I should test this PR hence @aulneau gave me the green signal to test them yesterday via our userx- tech channel.

yep this is the case, I had requested QA so that while i wrapped up things, i could fix anything else that might come up

aulneau:
closing in favor of https://github.com/blockstack/explorer/pull/493

Issue #472, opened 2021-06-29
[Microblocks] Add Microblocks page


fbwoolf:
Add a `Microblock` page which is linked from the block list:
https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2861%3A2245

Eshwari007:
Tested the Microblock page.

Issue #455, opened 2021-06-09
[Microblocks] Update txid page


kyranjamie:
**Individual transactions**
- Update to the title to show the state (pending, in microblock, or in anchor block): https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A1198
- Add component to show historical state changes: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A878 (this will be used in the transactions details view on the web wallet too)
- If included in anchor block, show number of anchor block confirmations since: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A878

Eshwari007:
Test passed for the first two use case.
Note:
If included in anchor block, show number of anchor block confirmations since: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A878 will not be implemented as it may be confusing for user

Issue #454, opened 2021-06-09
[Microblocks] Update block page with microblock row


kyranjamie:
**Block:**
https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2390%3A15241

fbwoolf:
Is this the right link? I don't see a microblock row?

fbwoolf:
Maybe this is the right link? https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A1477

Eshwari007:
The Block page in explorer is implemented as its in Figma
Howevere when anchor block is is selected,its not displaying as in the design.I have created a bug for the issue.https://github.com/blockstack/explorer/issues/503
Microblock page is tested and looking good.

Issue #453, opened 2021-06-09
[Microblocks] Update transaction list components


kyranjamie:
Used both on homepage and on the transaction pages

**Home**: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A2321
- The list of confirmed transactions is segmented by 'in microblock' and 'in anchor block'. There is no space to explain in line, so the info icons at either label should link to the docs which explain the difference, anchored to the respective section. 
- Blocks show microblocks and anchor blocks

**Transactions overview** https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A1735
- Same as the list of transactions on the home page
- You can see the interaction for the headers here: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A238
- Pending transaction: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A449

**Blocks overview**: https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Explorer?node-id=2847%3A1477
- Same as on homepage

fbwoolf:
I don't see any changes to the pending list? Also, not sure how to address the sticky header atm bc currently the entire page scrolls, not inside the actual list. Def can be changed but not current behavior.

Eshwari007:
Test passed both in Homepage and Transaction page

PR #482, opened 2021-07-13
feat: microblocks

fbwoolf:
## Description

This PR adds microblocks data to the the transactions list and the blocks list. In addition, it adds a new microblocks page.

image

image

For details refer to issue #453, #454, #455, #472

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other


src/features/blocks-list/microblock-list-item.tsx
```diff
@@ -0,0 +1,69 @@
...
+  blockTime: number;
+  hash: string;
+  index: number;
+  length: number;
+}> = React.memo(({ blockTime, hash, index, length, ...rest }) => {
+  const apiServer = useApiServer();
+  const getMicroblock = async () => {
+    return await fetchMicroblock(apiServer)(hash);
+  };
+  const { data: microblock } = useQuery('microblock', getMicroblock);
```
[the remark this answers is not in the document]
aulneau: Here is an example of a single block atom https://github.com/blockstack/explorer/blob/feat/cleanup-data/src/store/blocks.ts#L66

which you can then use like

```ts
const blockHash = '0xasdasdasda'
const [block] = useQueryAtom(blocksSingleState(blockHash));
```

src/store/currently-in-view.ts
```diff
@@ -98,6 +101,14 @@ export const contractInfoInViewState = atom(get => {
   }
 });
 
+export const currentlyInViewMicroblockHash = atom<string | undefined>(get => {
+  const inView = get(currentlyInViewState);
+  if (inView) {
+    if (inView.type === 'microblock') return inView.payload;
+  }
```
aulneau, line 108:
```suggestion
  if (inView && inView.type === 'microblock') return inView.payload;
```

src/features/blocks-list/microblock-list-item.tsx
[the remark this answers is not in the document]
aulneau: I see you made one! https://github.com/blockstack/explorer/pull/482/files#diff-45957df13ecef6be2b511ce3aa1ae417d5065df64738abc9335bdf2fd62194c7R32

aulneau (approved):
Nice this looks great!

Just a few comments, mostly about how there is an extra use of `react-query`, where we should only be using the atoms for any react-query functionality

src/features/blocks-list/microblock-list-item.tsx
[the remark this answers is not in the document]
fbwoolf: I did try this but couldn't get it to work ...but, I wasn't using `useQueryAtom()` so that might have been my problem. I'll try it!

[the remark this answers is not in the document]
aulneau: it's important that whatever the query is you're trying to use has been fetched by the page already, that's likely what caused it not to work

src/components/app-config.tsx
```diff
@@ -11,17 +12,21 @@ interface AppConfigProps {
   networkMode?: NetworkMode;
 }
 
+const queryClient = new QueryClient();
```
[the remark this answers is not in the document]
fbwoolf: I really just couldn't get anything to load the microblock item data in the list correctly unless I did this, but going to try your suggestion below. I must not have been using the right implementation of `useQueryAtom()`.

src/components/transaction-item.tsx
```diff
@@ -33,6 +33,8 @@ const getRelativeTimestamp = (tx: Transaction | MempoolTransaction) => {
   const date =
     typeof (tx as any).burn_block_time !== 'undefined' && (tx as any).burn_block_time !== -1
       ? toRelativeTime((tx as any).burn_block_time * 1000)
+      : (tx as any).burn_block_time === -1
+      ? 'Need parent burn block time...' // MICROBLOCK TODO: Replace with real data.
```
beguene, line 37: Maybe that's what you meant in your comment, but I think you could use 'parent_burn_block_time' here

src/components/transaction-item.tsx
```diff
@@ -188,21 +187,33 @@ const LargeVersion = React.memo(
...
               ) : null}
-
               {tx.tx_status !== 'pending' && tx.tx_status !== 'success' ? (
-                <Caption fontWeight="bold" mr="tight" as="span" color={color('feedback-error')}>
+                <Caption mr="6px" as="span" color={color('feedback-error')}>
                   Failed
                 </Caption>
               ) : null}
-              <Timestamp tx={tx} />
+              {tx.tx_status === 'success' && !!tx.microblock_hash ? (
```
beguene, line 202: I used tx.is_unanchored to detect microblocks, I think it's better because the tx will still have the attribute microblock_hash even when it is anchored
Example https://stacks-node-api-microblocks.testnet.stacks.co/extended/v1/address/STM6S3AESTK9NAYE3Z7RS00T11ER8JJCDNTKG711/transactions?limit=50

beguene (changes requested):

src/components/transaction-item.tsx
fbwoolf: Yep, I was just waiting for the new deployed version.

src/components/transaction-item.tsx
fbwoolf: It will have it but be an empty string (which is working), but I agree with you now that we have `is_unanchored` to work with. Good call. 👍

fbwoolf: I do not see the `is_unanchored` is ever `true`, do you? I don't think this is working?

fbwoolf:
@Eshwari007 I think this branch is going to fail to deploy bc of type checking with things still being published to beta versions. I think you might have to run it locally to test it. Just get on my branch and follow the readme to run it. You just have to set a few env variables, run `yarn && yarn dev`. Let me know if you can't get it running.

fbwoolf:
@beguene can you circle back and take a look at your requested changes next week while I'm out?

beguene (approved):

aulneau:
closing in favor of https://github.com/blockstack/explorer/pull/493

Issue #329, opened 2021-01-15
Emit metrics for Prometheus


wileyj:
**Describe the bug**
Port 9153 to emit metrics is no longer open

**Expected behavior**
port 9153 is open, and the path `/metics` reports TSDB data

wileyj:
bump

wileyj:
history:
because of how the explorer works, it can potentially track **millions** of paths for latency and other metrics. 
since we're using a pull based metrics model, every n seconds when we pull the metrics from the explorer the data we're pulling can be an extremely large dataset. 
what was happening was when metrics were enabled for the explorer, every n seconds the explorer latency would spike while the metrics were pulled. 
Not to mention the latency measurements are essentially unusable since each path is different. 

to resolve this for the API, @zone117x did some work to store similar paths into dedicated buckets. 
for example, all `/block/xxxx` requests would be stored in a `blocks` bucket, etc. 

I'll look for the PR in that repo that enabled this and link here shortly.

wileyj:
https://github.com/blockstack/stacks-blockchain-api/pull/502

https://github.com/blockstack/stacks-blockchain-api/blob/master/src/api/init.ts#L196-L226

markmhx:
Thanks @wileyj 🙏 Assigning to @aulneau so he can provide some input as well for anyone who might tackle this issue.

fbwoolf:
I can start looking at this if @aulneau can provide a bit more context?

aulneau:
Sure -- so we need to do a few things.

We are no longer using the custom server found here: https://github.com/blockstack/explorer/blob/main/server/index.js

We need to start using it again, and it needs to be updated in a few ways:
- remove the SSR caching that is included in it
- implement the same kind of middleware as in the api, [see that here](https://github.com/blockstack/stacks-blockchain-api/blob/master/src/api/init.ts#L78) and the one we have in the [server for the explorer](https://github.com/blockstack/explorer/blob/main/server/index.js#L97)
- implement a path regex like in the api [seen here](https://github.com/blockstack/stacks-blockchain-api/blob/master/src/api/init.ts#L196)
- make sure to read [this](https://github.com/blockstack/stacks-blockchain-api/issues/403) and [review this](https://github.com/blockstack/stacks-blockchain-api/pull/502)

Basically these changes make it so that all types of paths put their metrics into a single respective bucket, meaning rather than having a single metrics entry per address that is navigated to, we have all metrics combined into a single address bucket.

[...]

PR #491, opened 2021-07-22
fix: metrics

fbwoolf:
## Description

This PR fixes emitting metrics from the explorer by creating routes for a single address bucket rather than having a metric for every address navigated to by a user.

For details refer to issue #329 

Referenced PR:
https://github.com/blockstack/stacks-blockchain-api/pull/412

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other


fbwoolf:
@aulneau I'm not real sure if I'm on the right track with this so hoping for feedback ...how do I test this?

aulneau (commented):
This looks good to me, but I'd love a second look from @zone117x 

Thanks so much @fbwoolf!!!

server/index.js
```diff
@@ -14,104 +15,68 @@ const loggingMiddleware = require('./logging');
...
     '/_next/',
     expressStaticGzip('/static/', {
       enableBrotli: true,
     })
   );
 
+  let routes = [];
+
+  // Store all the registered express routes for usage with metrics reporting
+  routes = expressListEndpoints(app).map(endpoint => ({
```
zone117x, line 31: What is the value of `routes` after being set here? When I implemented this in the API, all the express routes had to be defined first before the `express-list-endpoints` lib could look them up. I'm not familiar with how routing works with the `next` js stuff.

server/index.js
```diff
@@ -14,104 +15,68 @@ const loggingMiddleware = require('./logging');
...
   if (!dev) {
+    const promMiddleware = createPrometheusMiddleware({
+      options: {
+        normalizePath: path => {
+          // Get the url pathname without a query string or fragment
+          // (note base url doesn't matter, but required by URL constructor)
+          let pathTemplate = new URL(path, 'http://x').pathname;
+          // Match request url to the Express route
+          for (const pathRegex of routes) {
+            if (pathRegex.regexp.test(pathTemplate)) {
```
zone117x, line 63: This can throw, see try/catch here https://github.com/blockstack/stacks-blockchain-api/blob/ab45a151f4136a1e8907befc1d6047035f6fd6f8/src/api/init.ts#L92-L107

server/index.js
```diff
@@ -14,104 +15,68 @@ const loggingMiddleware = require('./logging');
...
+    server.use(promMiddleware);
     createServer({ port: 9153 }).then(() => console.log(`@promster/server started on port 9153.`));
     server.use(loggingMiddleware());
   }
 
   server.get('/', (req, res) => {
     // since we don't use next's requestHandler, we lose compression, so we manually add it
     server.use(compression());
-    renderAndCache(app)(req, res, '/');
   });
```
aulneau, line 81: we dont need this I think

server/index.js
```diff
@@ -14,104 +15,68 @@ const loggingMiddleware = require('./logging');
...
 
   server.get('/*.js', (req, res, next) => {
```
aulneau, line 82: also this can go i think too

server/index.js
fbwoolf: Good point, I copied this over and in this case `app` should be `server` (just noticed that). I'm not sure about the express routes with nextjs either. I'll look into it. @aulneau any thoughts? Should I do something like this: https://github.com/blockstack/stacks-blockchain-api/blob/22e9a4f64f093fd9982dde570e18278b65e4b934/src/api/init.ts#L113

fbwoolf: I am not too familiar with express, so was looking at documentation ...was going to do something like:

`server.use(
    (() => {
      const router = express.Router();
      router.use('/txid', next());
      router.use('/microblock', next());
      router.use('/block', next());
      router.use('/address', next());
      return router;
    })()
  );`

fbwoolf:
@wileyj I am trying to figure out how to test the changes I'm making here, can you provide any direction?

server/index.js
```diff
@@ -14,141 +15,82 @@ const loggingMiddleware = require('./logging');
...
+  let routes = [];
 
-  server.get('/*.js', (req, res, next) => {
-    req.url = req.url + '.br';
-    res.set('Content-Encoding', 'br');
-    res.set('Content-Type', 'application/javascript; charset=UTF-8');
-    next();
-  });
+  // Store all the registered express routes for usage with metrics reporting
+  routes = expressListEndpoints(server).map(endpoint => ({
```
kyranjamie, line 42: Why declare routes on L39 and immediately redeclare it?

```
const routes = expressListEndpoints()....
```

wileyj:
> @wileyj I am trying to figure out how to test the changes I'm making here, can you provide any direction?

if you run this codebase locally (and everything is working as expected, of course) - it should open 2 ports: `3000` for the application, and `9153` for the metrics. 

from that, you can make a request for the metrics port like this:
`curl  localhost:9153/metrics` and you should see a response like this:
https://github.com/hirosystems/bitcoin#sample-metrics-output-via-curl

server/index.js
```diff
@@ -14,141 +15,82 @@ const loggingMiddleware = require('./logging');
...
+      options: {
+        normalizePath: path => {
+          // Get the url pathname without a query string or fragment
+          // (note base url doesn't matter, but required by URL constructor)
+          try {
+            let pathTemplate = new URL(path, 'http://x').pathname;
+            // Match request url to the Express route
+            for (const pathRegex of routes) {
+              if (pathRegex.regexp.test(pathTemplate)) {
+                pathTemplate = pathRegex.path;
```
zone117x, line 76: Seems like it would be useful to console.log to see if this code path is actually triggered when hitting on of the explorer pages. I'm not sure if the `expressListEndpoints` function is actually helping much here given the defined express routes above (i.e. not sure if `router.use('/txid', next())` generates regex that will get matched correctly).

Although, even if the auto-magic express route detection matching doesn't work, it does look like the manually defined paths and regexes in the `routes` array would work.

PR #493, opened 2021-07-27
MICROBLOCKS

aulneau:
This PR is the combination of https://github.com/blockstack/explorer/pull/469 and https://github.com/blockstack/explorer/pull/482

closes https://github.com/blockstack/explorer/issues/509


Eshwari007:
Tested both the PR which encompasses the ticket

src/components/contract-details.tsx
```diff
@@ -25,16 +26,16 @@ const PluralizedItem: React.FC<BoxProps & { array: any[]; label: string }> = ({
...
 
-export const ContractDetails = ({ contractId, contractInterface }) => {
+export const ContractDetails = ({ contractInterface, tx }) => {
   return (
     <Section flexShrink={0} mb="extra-loose" minWidth="200px" title="Contract details">
       <Box px="base">
         <Flex borderBottom={border()} py="base" justifyContent="space-between" alignItems="center">
           <Flex alignItems="center">
-            <ItemIcon size="48px" type="tx" txType="smart_contract" status="success" />
+            <ItemIcon size="48px" tx={tx} />
```
aulneau, line 35: @fbwoolf looking at this more -- this is not technically correct.

When on a contract-call page, we don't want to pass the contract-call transaction to the icon, we instead want to pass the parent contract information to this. maybe we can mock the tx object to be of the parent contract deploy, or fetch it (i think it's fetched already)

fbwoolf: Ok, I'll look at this. I thought we wanted the contract-call transaction here. Should be easy to change.

package.json
```diff
@@ -26,16 +26,16 @@
...
     "@promster/server": "6.0.2",
-    "@reach/tooltip": "0.15.2",
+    "@reach/tooltip": "0.15.3",
     "@react-aria/interactions": "3.5.0",
     "@react-spectrum/utils": "3.6.1",
     "@react-stately/toggle": "3.2.2",
-    "@stacks/auth": "2.0.0-beta.1",
-    "@stacks/blockchain-api-client": "0.62.1",
+    "@stacks/auth": "2.0.1-beta.1",
+    "@stacks/blockchain-api-client": "0.63.0-beta.0",
```
fbwoolf, line 34: The beta version on this can be removed, correct?

aulneau: that is correct

src/common/page-queries/microblock-hash.ts
```diff
@@ -0,0 +1,67 @@
...
+function getMicroblockPageCachedQueryProps(ctx: NextPageContext, queryClient: QueryClient) {
+  const microblockHash = getMicroblockHashFromCtx(ctx);
+  const queryKey = getMicroblocksQueryKey.single(microblockHash);
+  return getSingleCachedQueryData<Microblock>(queryKey, queryClient);
+}
+
+export const getMicroblockPageQueryProps = async (
+  ctx: NextPageContext,
+  queryClient: QueryClient
+  // TODO: Remove 'any' before deploy when types are updated in client package
```
beguene, line 26: @fbwoolf Shall this be updated ?

beguene (commented):
@fbwoolf Just one minor thing. I don't think this should block the release though.

src/common/page-queries/microblock-hash.ts
fbwoolf: @kyranjamie do you know why when I remove `any` I get this type error here?

image

aulneau: you'll need to cast it -- the types are different between the client and api-types libraries

Issue #539, opened 2021-09-13
Can not load regtest explorer from explorer.stacks.co 


314159265359879:
**Describe the bug**
Regtest not loading on explorer.stacks.co
![image](https://user-images.githubusercontent.com/33360391/133146254-e6838a97-0a86-41ca-bf11-f560c48cac65.png)
When I click on the red circle, I get the subsequent message lower middle of the page: You're viewing the mainnet explorer. It is not switching.
![image](https://user-images.githubusercontent.com/33360391/133146380-e3656484-7e6c-4676-b50f-53f9065f990b.png)

**To Reproduce**
1. Go to explorer.stacks.co
2. Click on Network (upper right corner)
3. Click on see screenshot: regtest
4. See error: You're viewing the mainnet explorer

**Expected behavior**
The regtest explorer should be shown, and I would like the ability to use the faucet.

**Screenshots**
See intro

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox
 - Version 

**Smartphone (please complete the following information):**
 - Version 92.0

**Additional context**
![image](https://user-images.githubusercontent.com/33360391/133146775-3a52f920-06f1-479b-98e1-765a4e048428.png)

314159265359879:
I think this is not yet solved the message has now changed to "...undefined explorer" and I am send to this URL when I am on the sandbox. When I am on the explorer's frontpage it appears to remain on mainnet when I click "regtest" in the network menu.

![image](https://user-images.githubusercontent.com/33360391/133784707-42deb419-5285-4e08-b0a9-6e2d147ceac5.png)

fbwoolf:
> I think this is not yet solved the message has now changed to "...undefined explorer" and I am send to this URL when I am on the sandbox. When I am on the explorer's frontpage it appears to remain on mainnet when I click "regtest" in the network menu.
> 
> ![image](https://user-images.githubusercontent.com/33360391/133784707-42deb419-5285-4e08-b0a9-6e2d147ceac5.png)

This has been merged, but not released into production yet. Let me release it this morning for you!

PR #540, opened 2021-09-13
fix: regtest network mode

fbwoolf:
## Description

This removes regtest as a default network from the explorer.

For details refer to issue #539 

Screen Shot 2021-09-15 at 1 52 06 PM

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Testing information

- Make sure switching to/from regtest network works properly with toast, url chain, and dropdown list


src/common/api/utils.ts
```diff
@@ -121,6 +129,8 @@ export const getChainTypeFromUrl = (networkUrl: string | undefined) => {
       return NetworkModes.Mainnet;
     case DEFAULT_TESTNET_SERVER:
       return NetworkModes.Testnet;
+    case DEFAULT_REGTEST_SERVER:
+      return NetworkModes.Regtest;
```
fbwoolf, line 133: This missing seemed to be the real issue ...the chain came in `undefined`.

kyranjamie (changes requested):
As I understand, there is no "regtest" chain mode. There is only `mainnet` and `testnet`. Regtest is of chain mode testnet (referred to as `ChainID` in stacks.js).

It's confusing because we refer to the networks with the same name, but there's distinction between the two which isn't reflected here.

Let's talk it through later, but I think we want some kind of map bridging network to chainid. Something like:

```
{
  mainnet: ChainId.Mainnet,
  testnet: ChainId.Testnet,
  regtest: ChainId.Testnet,
}
```

src/common/api/utils.ts
kyranjamie: Shouldn't we return `NetworkMode.Testnet` here?

fbwoolf:
> 
> Let's talk it through later, but I think we want some kind of map bridging network to chainid. Something like:
> 

I think maybe the explorer has it setup a bit differently client-side bc there is a `NetworkModes.Regtest`
https://github.com/blockstack/explorer/blob/a1598bc6eca597c32b3b428c8fdd77c09981a8da/src/common/types/network.ts

The first thing the app does is get the `apiServer`, so the regtest url, `DEFAULT_REGTEST_SERVER` is used to then get that `NetworkMode` and use it. I think what you are suggesting would be refactoring how things have already been setup rather than just fixing this issue?

aulneau:
One thing to note -- it's very likely that regtest will be decommissioned, I think we should probably remove it from the explorer.

lgalabru:
Should we just get rid of the Regtest option, and replace it with Devnet?

aulneau:
> Should we just get rid of the Regtest option, and replace it with Devnet?

Devnet is a local only version, right? local (non-https) urls will only work if the explorer is running locally, too.

markmhx:
Related: https://github.com/hirosystems/devops/issues/763

cc @CharlieC3

markmhx:
> Should we just get rid of the Regtest option, and replace it with Devnet?

I'd suggest this ☝️

fbwoolf:
@kyranjamie can you circle back and take a look through this one again?

kyranjamie (approved):
Circling back 🌀

LunarPunks:
Guys where are the settings in local explorer to make devnet the default,  i think this causing problems related to issue 647 and 648 ,  All actions divert to mainnet,  I need access to change this,  cannot do any local sandbox and explorer activities.

my setup is windows 10, clarinet, docker -  contracts deploy fine from clarinet --integrate but we cannot interact with them in sandbox,  because of the above mentioned problem.

Issue #58, opened 2020-04-10
Set up error tracking


kyranjamie:
On launch, we should have an error tracking tool in place. I'm familiar with [Sentry](https://sentry.io/welcome/) and would recommend it.

@agraebe Are you able to set up an account/billing?

diwakergupta:
I believe @hstove has already been using Sentry in another project. Please keep me in the loop on any setup here -- would prefer to have consolidated billing / accounting for all engineering related vendors.

agraebe:
@diwakergupta I haven't touched anything related to Sentry at Blockstack so far. Do we already have an account we could use and add a new project to it?

diwakergupta:
> @diwakergupta I haven't touched anything related to Sentry at Blockstack so far. Do we already have an account (with billing set up) we could use and add a new project to it?

We do, but I can't add anymore members without upgrading. I'm reaching out to see if they can give us a free upgrade (they have an OSS program). Stay tuned!

diwakergupta:
@aulneau never heard back from them, but we needn't block. We already have a Sentry account, I'm happy to add others and upgrade our plan if needed for now. Who all do I need to add?

kyranjamie:
@aulneau and I primarily, I think. We can make a separate errors channel in slack if anyone else wants to follow along.

aulneau:
@diwakergupta did you add @kyranjamie and myself? if not I'd love to be added to get this implemented sometime this week. Thanks!

diwakergupta:
@aulneau just did, back to you!

aulneau:
@diwakergupta thank you. I don't seem to have permissions to create a new project for the explorer, could you increase them for me?

diwakergupta:
> @diwakergupta thank you. I don't seem to have permissions to create a new project for the explorer, could you increase them for me?

done

markmhx:
I'm reopening this since it appears Sentry wasn't fully integrated with https://github.com/blockstack/explorer/pull/135 above.

PR #543, opened 2021-09-14
feat: add sentry

fbwoolf:
## Description

This PR adds basic Sentry setup in the explorer.

@aulneau @kyranjamie 
Are there specific options/configurations we want to include here?

For details refer to issue #58 

Screen Shot 2021-09-14 at 2 06 00 PM

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other


package.json
```diff
@@ -30,6 +30,7 @@
     "@react-aria/interactions": "3.5.0",
     "@react-spectrum/utils": "3.6.1",
     "@react-stately/toggle": "3.2.2",
+    "@sentry/nextjs": "^6.12.0",
```
kyranjamie, line 33: This should be a devDependency right? Should havet exact versions too, `-E` flag

sentry.client.config.js
```diff
@@ -0,0 +1,9 @@
+import * as Sentry from '@sentry/nextjs';
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
+
+Sentry.init({
+  dsn: SENTRY_DSN || 'https://388965ac235a42b990ef21332a471fc2@o204651.ingest.sentry.io/5236507',
```
kyranjamie, line 6: I don't think we should be committing this URL. Do we need a fallback versus using the env var?

kyranjamie (commented):
Looks good @fbwoolf—

It'd be great if there was a way to add sentry globally, rather than having to use the `withSentry` HOC on every page, but appreciate this is the way the guide does it.

Would recommend only injecting both the DSN URLs via environment variables. We don't want errors logged in our development environment, for example, so in these cases we just don't define said env vars.

sentry.client.config.js
fbwoolf: Yep, I'll move it to an env variable.

package.json
fbwoolf: Not sure on this, the closed PR before this didn't save it that way so went the same direction.

kyranjamie:
Oh yeah and does `withSentry` need to be used on every page, here it's only used on two.

aulneau:
yeah seems like it's only for api routes, see https://github.com/vercel/next.js/tree/canary/examples/with-sentry for another helpful implementation

kyranjamie:
interesting, are client side errors caught then? Don't we need a new Sentry-specific context/errorboundary?

aulneau:
My understanding is sentry rely's on the `_error.tsx` page that next.js has, see https://github.com/vercel/next.js/blob/canary/examples/with-sentry/pages/_error.js (not sure we have this yet, if not we'll need to add it)

fbwoolf:
@kyranjamie thought we could talk about how to handle the auth credentials with Github in our 1:1 tomorrow since you will need for implementation in the wallets too?

fbwoolf:
> @kyranjamie thought we could talk about how to handle the auth credentials with Github in our 1:1 tomorrow since you will need for implementation in the wallets too?

We forgot to talk about this, any ideas on how to handle the auth token to get past these build failures?

kyranjamie:
Let's go through it when you're awake. Another thought, do we need to make an exception for sentry in the CSP?

fbwoolf:
@kyranjamie and I ran through this together and one of us needs `Owner` permissions in Sentry to create an auth token for our organization. @diwakergupta can you either do this for us, or make one of us an `Owner`? We can then use it to setup a new repo secret with the auth token in Github to get past these build failures.

fbwoolf:
@kyranjamie not sure if this can help us get this setup correctly? (see section on Auth Tokens)
https://docs.sentry.io/product/integrations/integration-platform/#internal-integrations
Which links to: https://blog.sentry.io/2019/11/21/customize-your-sentry-workflow-a-sample-internal-integration

fbwoolf:
@kyranjamie did a deep dive into docs and resources online today but nothing I try works. I created repo secrets for the auth token and dsn, but didn't help. Devops doesn't have access to Sentry. My hunch is, we need to go the route of creating an org auth token (which I think is pretty simple), and create a Sentry Release Github Action? This will help us get the auth token working?
https://github.com/marketplace/actions/sentry-release
https://blog.sentry.io/2019/12/17/using-github-actions-to-create-sentry-releases

Why personal auth token might not be working?
https://docs.sentry.io/product/integrations/integration-platform/#how-is-the-sentry-integration-different-from-personal-auth-tokens

Another thought is, we could run the sentry-cli inside of docker and just use it to set the auth token there?
https://github.com/getsentry/sentry-cli#docker

If feeling better, maybe you'll have better luck in the AM?! The last commit I tried using `NEXT_PUBLIC_` syntax for the env variables and set the `authToken` directly in the webpack config, but not working. I know I can just set the `auth.token` in the `sentry.properties` but an env var doesn't seem to work there and I shouldn't push that info to Github. Also tried adding a curl cmd to the workflow but didn't work.

kyranjamie:
Thanks for investigating @fbwoolf.

I have a feeling we do have to use the regular auth tokens, rather than the org-wide ones. While it pushes you to create a proper integration, we just need the credentials, we don't have/need a webhook to point to.

I think one of the issues here is that, because we're building with docker, the environment variables aren't implicitly passed to the container, so the `sentry-cli` (used internally in the implicitly-used webpack plugin), doesn't have its auth options defined.

I've looked into the [docker build action](https://github.com/docker/build-push-action#inputs), for how to set these variables in the `docker build` cmd, but it's a little confusing. I've set the values for `build-args` to no avail, even the debug command. (Side rant, why does this action invent its own string-based key value pair format, rather than just use yaml....)

@CharlieC3 can we get your input here on how to proceed/configure the container with the environment variables

fbwoolf:
> 
> I have a feeling we do have to use the regular auth tokens, rather than the org-wide ones. While it pushes you to create a proper integration, we just need the credentials, we don't have/need a webhook to point to.
> 

Ok, wasn't confident on this either and couldn't test it w/ the current permissions I have. I wasn't sure if the webhook was required. Docs aren't great.

> I think one of the issues here is that, because we're building with docker, the environment variables aren't implicitly passed to the container, so the `sentry-cli` (used internally in the implicitly-used webpack plugin), doesn't have its auth options defined.
> 

Agree ...at least I'm learning quite a bit about Github workflows through this. Haha.

> I've looked into the [docker build action](https://github.com/docker/build-push-action#inputs), for how to set these variables in the `docker build` cmd, but it's a little confusing. I've set the values for `build-args` to no avail, even the debug command. (Side rant, why does this action invent its own string-based key value pair format, rather than just use yaml....)
> 

I looked at this too but didn't get anywhere, so hopefully Charlie can shed some light.

Thanks for helping me look into it!

Issue #545, opened 2021-09-15
Selection of new network in explorer points to mainnet


Eshwari007:
User added a new network via Web wallet as it wasnt able to do so via the explorer:https://github.com/blockstack/explorer/issues/532?
Upon navigating to explorer,user however can view the new network but when user selects the new network,it points to mainnet
<img width="400" alt="Screen Shot 2021-09-15 at 10 30 24 AM" src="https://user-images.githubusercontent.com/38872890/133452885-d85f6f15-7383-4107-ae00-ac3d8d63afe3.png">

fbwoolf:
> User added a new network via Web wallet as it wasnt able to do so via the explorer:#532?
> Upon navigating to explorer,user however can view the new network but when user selects the new network,it points to mainnet

@Eshwari007 I don't see a new network in the list you attached as the image? Were you selecting `regtest`? I'll try to repro this but we do have a PR open for removing regtest from the network list and it was a known issue that selecting it wasn't working.

Eshwari007:
<img width="415" alt="Screen Shot 2021-09-15 at 5 21 57 PM" src="https://user-images.githubusercontent.com/38872890/133511320-4483f978-480a-4561-9f9e-8593e57dc0df.png">
@fbwoolf  Attaching the network I selected which was directed to mainnet upon selection

fbwoolf:
> @fbwoolf Attaching the network I selected which was directed to mainnet upon selection

That was the original image posted with the issue, but the list you are showing is the default list of networks. I don't see a custom network there?

Eshwari007:
@fbwoolf I added the stacks.co api.regtest via WW.I only had mainnet and testnet on my list of networks.
Upon selecting the api regtest on explorer,the direction went to mainnet

Eshwari007:
The above was in my WW,interestingly am not able to view the newly created network in explorer
<img width="425" alt="Screen Shot 2021-09-15 at 5 30 06 PM" src="https://user-images.githubusercontent.com/38872890/133512407-7b9764d8-0867-40a7-8d9b-f11e9e3633ae.png">

fbwoolf:
This is expected behavior. Adding a network to the WW would only add it there. You would have to add it separately in the explorer.

fbwoolf:
I will test this anyway and make sure it works now that I fixed the `Add a network` modal.

fbwoolf:
This is not working, so good it was logged as an issue! Thx for helping to clarify.

PR #547, opened 2021-09-16
fix: custom network list

fbwoolf:
## Description

This PR is branched off fixing the `Add a network` modal. Once the custom network was added, it wasn't appearing in the network list. This fixes that issue. The expected behavior is that the chain id and toast will still be either `mainnet` or `testnet`, but if the user last used a custom saved network it will stay on that network and display it as selected in the network list.

*I need to merge this into the base branch before merging that PR into main and releasing.

For details refer to issue #545 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Testing information

- Add a new network via the modal
- It should stay on that network when added
- Network list will show the custom network as selected
- Chain id in query params and toast will still read either `mainnet` or `testnet`
- Removing the network should return to the default network (mainnet)


src/common/api/utils.ts
```diff
@@ -32,55 +39,45 @@ export const getServerSideApiServer = async (ctx: NextPageContext) => {
...
-        maxAge: 30 * 24 * 60 * 60,
-        path: '/',
-      });
-    } else if (chain === 'testnet') {
-      apiServer = DEFAULT_NETWORK_LIST[DEFAULT_TESTNET_INDEX]?.url;
-      nookies.set(ctx, NETWORK_CURRENT_INDEX_COOKIE, JSON.stringify(DEFAULT_TESTNET_INDEX), {
-        maxAge: 30 * 24 * 60 * 60,
-        path: '/',
-      });
+  if (INDEX === 0 || INDEX === 1) {
```
fbwoolf, line 46: I will change these to use existing constants.

src/common/api/utils.ts
```diff
@@ -32,55 +39,45 @@ export const getServerSideApiServer = async (ctx: NextPageContext) => {
...
+    if (chain) {
+      if (chain === 'mainnet') {
```
fbwoolf, line 48: Same with these (line 53 too).

kyranjamie (approved):

Issue #556, opened 2021-09-29
Label orphaned microblocks transactions accurately


gigawatson:
This transaction, with a nonce of `376`, was confirmed 6 hours ago: https://explorer.stacks.co/txid/caf14aa6aae66c121e69bd90e37321b8e56b67e4c478de27b6847d8f6f62d8ed?chain=mainnet

However, the last confirmed transaction for that address under [its history](https://explorer.stacks.co/address/SP19WSDJWTH4CW3YG554XS5CAXJJGAN83P8CFZ4K1?chain=mainnet) is nonce `375`, which strangely shows as confirmed 2 hours _after_ nonce `376`.

Meanwhile, a transaction with a nonce `377` is stuck pending.

May be related to blockstack/stacks-wallet-web#1723 which was recently closed.

PR #560, opened 2021-10-02
fix: orphaned microblock transaction status

fbwoolf:
## Description

This PR refactors tx status and fixes orphaned microblock transactions to have a status of `Failed`. Previously, those transactions were being labeled as successful in the UI so it appeared there were two successful transactions with the same nonce.

For details refer to issue #556 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -3,6 +3,7 @@ import getConfig from 'next/config';
 import packageJson from '../../package.json';
 
 import { Transaction } from '@stacks/stacks-blockchain-api-types';
+import { TxStatus } from './types/tx';
 
 const { publicRuntimeConfig } = getConfig();
 
@@ -134,3 +135,10 @@ export const TransactionType = {
   COINBASE: 'coinbase' as Transaction['tx_type'],
   POISON_MICROBLOCK: 'poison_microblock' as Transaction['tx_type'],
 } as const;
+
+export const TransactionStatus = {
+  PENDING: 'pending' as TxStatus,
+  SUCCESS_ANCHOR_BLOCK: 'success_anchor_block' as TxStatus,
+  SUCCESS_MICROBLOCK: 'success_microblock' as TxStatus,
+  FAILED: 'failed' as TxStatus,
+};
--- a/src/common/hooks/use-transaction-status.ts
+++ b/src/common/hooks/use-transaction-status.ts
@@ -1,0 +1,34 @@
+import React, { useEffect, useState } from 'react';
+import { MempoolTransaction, Transaction } from '@stacks/stacks-blockchain-api-types';
+import { TxStatus } from '@common/types/tx';
+import { TransactionStatus } from '@common/constants';
+
+export const useTransactionStatus = (
+  tx: Transaction | MempoolTransaction | undefined
+): TxStatus | undefined => {
+  const [txStatus, setTxStatus] = useState<TxStatus>();
+
+  // Used to construct tx status feedback in the UI
+  useEffect(() => {
+    if (tx?.tx_status === 'pending') {
+      setTxStatus(TransactionStatus.PENDING);
+    } else if (tx?.tx_status === 'success' && !!tx.is_unanchored) {
+      setTxStatus(TransactionStatus.SUCCESS_MICROBLOCK);
+    } else if (tx?.tx_status === 'success' && !tx.is_unanchored) {
+      if (!tx.microblock_canonical) {
+        setTxStatus(TransactionStatus.FAILED);
+      } else {
+        setTxStatus(TransactionStatus.SUCCESS_ANCHOR_BLOCK);
+      }
+    } else if (
+      tx?.tx_status === 'abort_by_response' ||
+      tx?.tx_status === 'abort_by_post_condition'
+    ) {
+      setTxStatus(TransactionStatus.FAILED);
+    } else {
+      setTxStatus(undefined);
+    }
+  });
+
+  return txStatus;
+};
--- a/src/components/function-summary/result.tsx
+++ b/src/components/function-summary/result.tsx
@@ -5,11 +5,20 @@ import { Caption, Pre } from '@components/typography';
 import { border } from '@common/utils';
 import { FunctionSummaryClarityValue } from '@components/function-summary/value';
 import { IconAlertTriangle, IconCircleCheck } from '@tabler/icons';
+import { TransactionStatus } from '@common/constants';
 
-export const FunctionSummaryResult = ({ result }: { result: Transaction['tx_result'] }) => {
+export const FunctionSummaryResult = ({
+  result,
+  txStatus,
+}: {
+  result: Transaction['tx_result'];
+  txStatus: string | undefined;
+}) => {
   if (!result) return null;
-  const { success, type, value } = cvToJSON(hexToCV(result.hex));
-
+  const { type, value } = cvToJSON(hexToCV(result.hex));
+  const isSuccess =
+    txStatus === TransactionStatus.SUCCESS_ANCHOR_BLOCK ||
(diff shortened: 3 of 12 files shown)


src/common/constants.ts
```diff
@@ -134,3 +135,10 @@ export const TransactionType = {
   COINBASE: 'coinbase' as Transaction['tx_type'],
   POISON_MICROBLOCK: 'poison_microblock' as Transaction['tx_type'],
 } as const;
+
+export const TransactionStatus = {
+  PENDING: 'pending' as TxStatus,
+  SUCCESS_ANCHOR_BLOCK: 'success_anchor_block' as TxStatus,
+  SUCCESS_MICROBLOCK: 'success_microblock' as TxStatus,
+  FAILED: 'failed' as TxStatus,
+};
```
kyranjamie, line 144: I think you can avoid these duplicate type casts if you type as

```suggestion
export const TransactionStatus: Record<string, TxStatus> = {
  PENDING: 'pending',
  SUCCESS_ANCHOR_BLOCK: 'success_anchor_block',
  SUCCESS_MICROBLOCK: 'success_microblock',
  FAILED: 'failed',
};
```

src/common/hooks/use-transaction-status.ts
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from 'react';
+import { MempoolTransaction, Transaction } from '@stacks/stacks-blockchain-api-types';
+import { TxStatus } from '@common/types/tx';
+import { TransactionStatus } from '@common/constants';
+
+export const useTransactionStatus = (
```
kyranjamie, line 6: Does this need to be a hook? It isn't reliant on any external state. So it can just be a function that accepts a transaction and returns a value.

src/components/function-summary/result.tsx
```diff
@@ -5,11 +5,20 @@ import { Caption, Pre } from '@components/typography';
...
+import { TransactionStatus } from '@common/constants';
 
-export const FunctionSummaryResult = ({ result }: { result: Transaction['tx_result'] }) => {
+export const FunctionSummaryResult = ({
+  result,
+  txStatus,
+}: {
+  result: Transaction['tx_result'];
+  txStatus: string | undefined;
+}) => {
```
kyranjamie, line 16: As we're touching this file, can we set props with an interface:

```suggestion
interface FunctionSummaryResultProps {
  result: Transaction['tx_result'];
  txStatus: string | undefined;
}
export const FunctionSummaryResult = ({ result, txStatus }: FunctionSummaryResultProps) => {
```

src/components/item-icon.tsx
```diff
@@ -36,8 +38,8 @@ const ItemCircle: React.FC<GridProps> = props => (
   />
 );
 
-const StatusBubble: React.FC<any> = ({ tx }) => {
-  if (tx?.tx_status === 'pending') {
+const StatusBubble: React.FC<any> = ({ txStatus }) => {
```
kyranjamie, line 41: Also see this was done as `any` previously, but this'd be a good time to type properly

src/common/hooks/use-transaction-status.ts
fbwoolf: I don't think using a custom hook is dependent on relying on external state? It is extracting out setting state into a reusable hook? I did forget the dependency array though. We seem to use custom hooks in a lot of places that don't rely on external state. Not right?

src/common/constants.ts
fbwoolf: Thx for the suggestion, I wasn't sure how to do this.

src/common/hooks/use-transaction-status.ts
kyranjamie: Sorry for essay, hope this is a helpful explanation.

You're right, hooks aren't only reliant on using external state, but they enable access to a handful of features right. Access to state methods that persist between renders, `useEffect` to manage code dependent on changing props, other hooks etc.

It's not clear, based on this implementation, why we need any of these. The `useState` is updated on every change, but why not just return the value immediately?

Consider if we need to reuse this code elsewhere: directly in an atom, within another function etc. We're unable to because of the decision to make it a hook, as they only work in components/other hooks. 
Sometimes we need hooks of course, for one of the reasons stated above, but the core logic of it is usually better described as a pure function. 

Here's a simplified implementation:

```ts
export function getTransactionStatus(tx: Transaction | MempoolTransaction) {
  if (tx?.tx_status === 'pending') {
    return TransactionStatus.PENDING;
  } else if (tx?.tx_status === 'success' && tx.is_unanchored) { // <-- note, no need for `!!` here, as the value is already a boolean
    return TransactionStatus.SUCCESS_MICROBLOCK;
  } else if (tx?.tx_status === 'success' && !tx.is_unanchored) {
    if (!tx.microblock_canonical) {
      return TransactionStatus.FAILED;
    } else {
      return TransactionStatus.SUCCESS_ANCHOR_BLOCK;
    }
  } else if (tx?.tx_status === 'abort_by_response' || tx?.tx_status === 'abort_by_post_condition') {
    return TransactionStatus.FAILED;
  }
}
```

Another important change I've made is to not bake in the `if undefined, return undefined` logic. This is additional functionality that 1) breaks the single responsibility principle, 2) makes an assumption the tx can sometimes be `undefined`. See it as a coupling of getting the tx status _and_ allowing it to be undefined. It's almost always better to leave the handling of this to whatever code invokes the function, rather than the fn itself. 

It also leverages the TypeScript compiler. If I use this again, and I pass it a value that might be undefined, the compiler will tell me. It's a silent, unobvious feature if it accepts `undefined`.

If you want to get #functional. We could make a wrapper method that composes this functionality. (Though this isn't necessary in this example).

```js
 // untyped example
function acceptUndefined (fn) {
  return (value) => value === undefined ? undefined : fn(value);
}
const newFunctionThatAcceptUndefined = acceptUndefined(getTransactionStatus)
```

src/components/item-icon.tsx
```diff
@@ -90,13 +92,15 @@ export const ItemIcon = React.memo(
     type: 'tx' | 'microblock' | 'block' | 'principal';
     tx?: Transaction | MempoolTransaction;
   } & GridProps) => {
+    const txStatus = useTransactionStatus(tx);
```
kyranjamie, line 95: Here, where we're using the fn, we may indeed wish to prevent multiple renders of the fn, but we can do this by memoizing the function.

```suggestion
    const txStatus = useMemo(() => getTransactionStatus(tx), [tx])
```

src/components/transaction-title.tsx
```diff
@@ -70,21 +71,14 @@ export const getTxTitle = (transaction: Transaction | MempoolTransaction) => {
...
 export const TransactionTitle = ({ contractName, tx, ...rest }: TitleProps & StackProps) => {
-  let txStatus: TxStatus;
-  if (tx.tx_status === 'success' && !!tx.is_unanchored) {
-    txStatus = 'success_microblock';
-  } else if (tx.tx_status === 'success' && !tx.is_unanchored) {
-    txStatus = 'success_anchor_block';
-  } else {
-    txStatus = tx.tx_status;
-  }
+  const txStatus = useTransactionStatus(tx);
```
kyranjamie, line 74: If you don't want to repeat the `useMemo` wrapper of the function, you _could_ make a hook the contains just the 
`useMemo(() => getTransactionStatus(tx), [tx])` called `useMemoizedTxStatus` or something.

But personally I don't think this abstraction is super important.

src/common/hooks/use-transaction-status.ts
fbwoolf: I appreciate the explanation. I def get your point now, thanks for taking the time to write it up!

fbwoolf:
@beguene would you mind looking at this in the morning and approving it if can? Thanks!

PR #563, opened 2021-10-09
feat: add segment

fbwoolf:
This add the initial install/setup of Segment in the explorer. Page views and event tracking are successfully logging to the web portal.

*Note: there is currently a rerender issue where pages/events are being recorded multiple times.


fbwoolf:
> How might we create a wrapper around `global.analytics` that each file can import?
> 
> We should use some kind of wrapping function. This way we can type it more easily, extend it with additional functionality etc (not possible if we call it directly)

This can be reviewed again.

Issue #569, opened 2021-10-15
Display clearly when txs are micro-orphaned 


andresgalante:
There is a report in discord of discrepancies in some transactions that I think we should investigate further:

https://discord.com/channels/621759717756370964/745197302255321108/898646581883068516

> Hi there,
I think there is a bug where the APIs that run the Explorer & Hiro wallet are out of sync with the blockchain.
I have a few examples of this, both of "disappearing" NFTs as well as STX balance that is outdated.
>
> I'll just show 1 example now, can share others later.
>
> 1. Buy Phases of Satoshi 94
>This address bought Phases of Satoshi `#94` about 8 days ago, but it does not show up in their account
https://explorer.stacks.co/address/SP3ZTTRW93D37AAEYQPAR1QQ8H8CQ7N1QCQ6KXW6V?chain=mainnet (also notice nonce 1 btw, while no TX has been done)
>
>The TX to purchase `#94` is here: https://explorer.stacks.co/txid/0xd82d04c0f832cf1496306a9b65f93b5ed44bc463abc307b893ca3499235f1664?chain=mainnet
>You can see it is successful and that the postcondition states that STX and an NFT should be transferred.
>
>Why is the transaction not showing in the Transactions tab on their profile (first link)? The nonce is 1 but no TX had been done (only 1 STX receival which does not count as a nonce).. so it seems the transaction has happened (since the nonce has awarenes of it?) but the TX does not show in the list, nor is the NFT showing in the Collections tab.
>
>This seems to be an issue with the Explorer/Wallet API, no? Can someone check this? 

cc @rafaelcr

PR #570, opened 2021-10-19
fix: display of non-canonical transactions

fbwoolf:
## Description

This PR fixes displaying both `canonical: false` and `microblock_canonical: false` transactions with a tx status of `Non-cononical (orphaned)` in the tx header and the changes the `Non-canonical` summary message to, `Transaction has been orphaned by the canonical chain. It is in a non-canonical fork.` This also updates the information icon link to the correct SIP location for more information: https://github.com/stacksgov/sips/blob/main/sips/sip-001/sip-001-burn-election.md#committing-to-a-chain-tip. The result status will now show the execution result correctly.

Issue #569 

Screen Shot 2021-10-19 at 10 46 23 AM

Screen Shot 2021-10-19 at 10 46 50 AM

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -143,5 +143,6 @@ export const TransactionStatus: Record<string, TxStatus> = {
   PENDING: 'pending',
   SUCCESS_ANCHOR_BLOCK: 'success_anchor_block',
   SUCCESS_MICROBLOCK: 'success_microblock',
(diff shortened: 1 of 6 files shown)


fbwoolf:
> 
> I think there should be a more pronounced "warning" looking display that has its own row.

Thanks for the feedback and opening the issue ...definitely seems like a good improvement to make. 👍

Issue #571, opened 2021-10-19
Segment not logging in production


fbwoolf:
Segment was working in dev env, but once deployed I don't see page views or events logging. I added a repo secret key for the  write_key, so I'm not sure why it wouldn't be logging. Need to look into it.

fbwoolf:
@CharlieC3 I got it working! Finally.

PR #582, opened 2021-10-29
fix: segment not logging in prod

fbwoolf:
This PR should fix Segment in production 🤞 . Moved rendering the snippet into the `_document`. Now able to use the node env var w/out the `NEXT_PUBLIC_` syntax which would expose the write key in the browser.

Issue #571


fbwoolf:
Build failure here is an active issue in webpack and in next tracked here:
https://github.com/webpack/webpack/issues/14532
https://github.com/vercel/next.js/issues/30078

Last suggestion is to downgrade node to 16 in the Dockerfile.

fbwoolf:
@CharlieC3 I got the build passing now if you want to test/approve.

CharlieC3:
@fbwoolf I believe you're able to deploy to staging. Once you do we can see if it works 🤞 
https://github.com/hirosystems/explorer/actions/runs/1400138068

fbwoolf:
> @fbwoolf I believe you're able to deploy to staging. Once you do we can see if it works 🤞 https://github.com/hirosystems/explorer/actions/runs/1400138068

Nope, not logging and Segment script is not in the header. Anything that would be different w/ staging vs production?

CharlieC3:
@fbwoolf I found an issue with the previous deployment caused by the GH org change. It should be fixed now for future staging deployments, but I had to manually deploy this branch. We'll encounter this issue for prod as well, so I may need to perform the next prod deployment.

When you get a chance can you try it again? `fix-segment-prod` is running _without_ the `SEGMENT_WRITE_KEY` env var set.

fbwoolf:
> 
> When you get a chance can you try it again? `fix-segment-prod` is running _without_ the `SEGMENT_WRITE_KEY` env var set.

Didn't see this and was waiting to deploy again with a change I made, so your fix is likely what got it working I guess. Funny. Anyway, it is working now.

CharlieC3:
Still a great success! Also I think the next deployment to `prod` should work now without intervention. I had to push up `1.17.2` to our new `hirosystems` Docker org .

CharlieC3 (approved):
:shipit:

fbwoolf:
Thanks, @CharlieC3 for all your help!

pkellner:
Thanks for posting this fix. I guess I need to start locking my alpine version to make sure my builds don't crash down the road for infrequently built apps.

PR #588, opened 2021-11-20
feat: BNS support

johnletey:
- [x] Address page
- [x] Transaction page
- [x] Search

<details>
  <summary>PoW</summary>
  <video src="https://user-images.githubusercontent.com/62398724/142733205-ef1e642e-5048-481d-acfa-d936b0e35ebe.mov" />
</details>

### DIFF
--- a/src/common/page-queries/account.ts
+++ b/src/common/page-queries/account.ts
@@ -6,14 +6,30 @@ import { DEFAULT_LIST_LIMIT } from '@common/constants';
 import { MempoolTransactionListResponse } from '@stacks/stacks-blockchain-api-types';
 import { InfoQueryKeys } from '@store/info';
 
-export function getPrincipalFromCtx(ctx: NextPageContext) {
+export async function getPrincipalFromCtx(ctx: NextPageContext) {
   const { query } = ctx;
-  const principal: string = query?.principal as string;
+  let principal: string = query?.principal as string;
+
+  // TODO: Is this the correct place for this?
+  if (principal.endsWith('.stx')) {
+    const { bnsApi } = await getApiClients(ctx);
+
+    try {
+      const res = await bnsApi.getNameInfo({
+        name: principal,
+      });
+
+      principal = res.address;
+    } catch {
+      principal = '';
+    }
+  }
+
   return principal;
 }
 
 export const getAccountPageQueries: GetQueries = async ctx => {
-  const principal = getPrincipalFromCtx(ctx);
+  const principal = await getPrincipalFromCtx(ctx);
   const { accountsApi, bnsApi, transactionsApi, infoApi } = await getApiClients(ctx);
 
   return [
--- a/src/common/page-queries/account.ts
+++ b/src/common/page-queries/account.ts
@@ -14,10 +14,21 @@ export function getPrincipalFromCtx(ctx: NextPageContext) {
 
 export const getAccountPageQueries: GetQueries = async ctx => {
   const principal = getPrincipalFromCtx(ctx);
-  const { accountsApi, transactionsApi, infoApi } = await getApiClients(ctx);
+  const { accountsApi, bnsApi, transactionsApi, infoApi } = await getApiClients(ctx);
 
   return [
     [InfoQueryKeys.INFO, async () => infoApi.getCoreApiInfo()],
+    [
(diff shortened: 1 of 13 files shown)


markmhx:
Thanks for the contribution! This looks fantastic. We'll code review soon 🙏

johnletey:
You're welcome @markmhx!

There's still two things I need to fix on my end 👀

1. Searching sometimes goes into this infinite loading state (not sure why)
2. When loading the address page, sometimes get an error about no initial data for the name

src/common/page-queries/account.ts
```diff
@@ -6,18 +6,47 @@ import { DEFAULT_LIST_LIMIT } from '@common/constants';
...
+      const res = await bnsApi.getNameInfo({
+        name: principal,
+      });
+
+      principal = res.address;
+    } catch {
+      principal = '';
+    }
+  }
+
```
kyranjamie, line 29: Perhaps here we could check for valid STX addresses, and return early when it matches, avoiding unnecessary calls

johnletey: @kyranjamie So I had it implemented previously as a simple `.endsWith` check, happy to switch back to that method.

But this approach allows us to resolve `.btc` addresses etc

kyranjamie: Yes, I saw this. As I understand, this fn is pulling the query info from the request, and it _could_ contain a valid STX address.

If this is the case, it's a wasted effort to request all this bns info, right? Couldn't we just return early? We could use the `c32check` lib to validate the input. There's already a helper for this, actually, `validateStacksAddress`

src/common/page-queries/transactions.ts
```diff
@@ -5,16 +5,43 @@ import { DEFAULT_LIST_LIMIT } from '@common/constants';
...
+        let transactions = (await transactionsApi.getTransactionList({
           limit: DEFAULT_LIST_LIMIT,
           offset: 0,
           type: DEFAULT_TX_FILTER_TYPES,
         })) as TransactionsListResponse;
+
+        for (let i = 0; i < transactions.results.length; i++) {
+          let res;
+
+          res = await bnsApi.getNamesOwnedByAddress({
```
kyranjamie, line 22:
```suggestion
          const res = await bnsApi.getNamesOwnedByAddress({
```

src/common/page-queries/transactions.ts
```diff
@@ -5,16 +5,43 @@ import { DEFAULT_LIST_LIMIT } from '@common/constants';
...
+            address: transactions.results[i].sender_address,
+            blockchain: 'stacks',
+          });
+          // @ts-ignore
```
kyranjamie, line 26: Where possible, we try to avoid ts-ignore checks. 

Here, the type errors are because `transactions.results[i]` is type Transaction, but `sender_address` only exists on `TokenTransferTransaction`. We can satisfy the compiler by checking the tx has `tx_type === 'token_transfer'`

kyranjamie (commented):
Thanks for the contribution @johnletey, looks great.

Added a few comments. Also encourage you to run `yarn lint`, as our CI doesn't run on forks.

johnletey:
Thanks for you feedback @kyranjamie, refactored the code!

Ran `yarn lint`, and there's quite a few warnings and errors 😅
Happy to work through all 271 problems, but quite a few aren't introduced by my code.

Again not an issue, happy to resolve all of them!

kyranjamie:
Hey @johnletey, sorry for not getting back to you. You absolutely don't need to address linting issues not related to this PR, that'd be a huge ask. 

The main reluctance to merge this PR as is would be the use of `// @ts-ignore`. We really need to avoid this, otherwise we may later introduce bugs that the TypeScript compiler might catch.

If you have time to address these, that'd be great, otherwise a Hiro engineer will eventually get to it

johnletey:
Hey @kyranjamie @andresgalante 👋🏻 Completely understand about `@ts-ignore`, will get the code updated soon!

andresgalante:
thanks for your contribution, I'll close this PR for now since it's stagnated. We can re open it later as we work on it.

andresgalante:
@He1DAr would you mind taking a look at this PR and assessing if there is a path for us to include this feature by improving this Pr?

Issue #604, opened 2021-12-17
parseFloat makes quantities / referenced amounts confusing to end users and integrators


wbnns:
**Describe the bug**
The use of parseFloat in the explorer is confusing to integrators and end users because it gives an incorrect perception of the amount being transferred, and the balance of holdings.

Reference # 1, e.g. balance of USDA is not $330,891,881.00:

<img width="340" alt="image" src="https://user-images.githubusercontent.com/1130872/146533798-9329fd4d-63e2-4447-b834-42999891810d.png">

Link to explorer:
https://explorer.stacks.co/address/SPWVN0HJ5RJJWXPAP5JFNZ9Q0V68TVSS1WSG45X6?chain=mainnet

Reference # 2, e.g. this is not a transfer of $11B USD

<img width="768" alt="image" src="https://user-images.githubusercontent.com/1130872/146534327-64ef609e-1447-4358-9e17-b1989530e7d3.png">

Link to explorer:
https://explorer.stacks.co/txid/0x8e8773bf0693a9ba7543eb2bef58634e91d8fc49cc0c2a5132c391843cf7acc6?chain=mainnet

**Expected behavior**
Explorer should take decimals into account for token (`get-decimals`).

philipdesmedt:
+1 - seen this confusion often with Arkadiko users (for all FTs, including USDA and DIKO)

wbnns:
Related: #402

fbwoolf:
Related: #506

wbnns:
@fbwoolf Thanks!

wbnns:
@fbwoolf :raised_hands:

fbwoolf:
Hoping to release this to production today! 🙌

fbwoolf:
@wbnns this should be live now.

Issue #506, opened 2021-08-05
Display decimal points as indicated by FT standard (SIP 010)


timstackblock:
There seems to be decimal issues with the Stella token it displays as 3.703703671 in the web wallet and displays 3,703,703,671.00 in the Explorer.

Since the wallet displays the wrong total users can send more than they have in the wallet example my wallet says 3.70 tokens but I can send 30000000.

Please see the address  here https://explorer.stacks.co/address/ST297VG59W96DPGBT13SGD542QE1XS954X4ZSWW2J?chain=testnet

<img width="387" alt="Screen Shot 2021-08-05 at 4 39 24 PM" src="https://user-images.githubusercontent.com/49165468/128418916-18d01c21-cbf7-4618-aa5b-a755dd21d106.png">

<img width="397" alt="Screen Shot 2021-08-05 at 4 39 06 PM" src="https://user-images.githubusercontent.com/49165468/128418931-b4b89b48-28dc-4db3-818c-0485daf52867.png">

<img width="448" alt="Screen Shot 2021-08-05 at 5 10 46 PM" src="https://user-images.githubusercontent.com/49165468/128421906-1ba262e3-6cdc-4220-8595-559af3373b52.png">

<img width="514" alt="Screen Shot 2021-08-05 at 5 10 51 PM" src="https://user-images.githubusercontent.com/49165468/128421896-5d7bbde6-4b11-4919-9787-d6b574985c64.png">

markmhx:
[Reported as a need for Friedger Pool in Discord.](https://discord.com/channels/621759717756370964/824628945982849055/891014837763735572)

markmhx:
I'm moving this to the backlog since it keeps coming up as confusing for folks.

Issue #402, opened 2021-03-04
Tokens are displayed with decimals


friedger:
**Describe the bug**
In the Holdings card, tokens are displayed with decimals even though they have 0 decimals.

![Screenshot from 2021-03-04 10-34-30](https://user-images.githubusercontent.com/1449049/109943523-92efcf00-7cd5-11eb-8bfe-00e765dd7593.png)

fbwoolf:
Related: #506

PR #612, opened 2021-12-31
feat: add sip 10 decimal support

fbwoolf:
This isn't the most DRY implementation, but bc of how these pages/components are built, this might be the best solution until we can give the explorer more attention and remove query atoms like we have done in the web wallet. The ft/nft tokens really need to be handled separately so that the fetch for data isn't done conditionally (which breaks the rules of hooks, so can't do it). This does resolve issues with missing ft decimals for now.

Issues: #402, #506, and #604

Screen Shot 2021-12-28 at 12 29 45 PM

Screen Shot 2021-12-28 at 2 44 16 PM

Screen Shot 2021-12-28 at 3 58 06 PM

### DIFF
--- a/src/common/utils.tsx
+++ b/src/common/utils.tsx
@@ -16,9 +16,10 @@ import relativeTime from 'dayjs/plugin/relativeTime';
 import { ContractCallTxs } from '@common/types/tx';
 import { Text } from '@components/typography';
 import { IconArrowLeft } from '@tabler/icons';
-import { TESTNET_CHAIN_ID } from '@common/constants';
+import { STX_DECIMALS, TESTNET_CHAIN_ID } from '@common/constants';
 import { NetworkMode, NetworkModes } from '@common/types/network';
 import { NextPageContext } from 'next';
+import BigNumber from 'bignumber.js';
 
 dayjs.extend(relativeTime);
 
@@ -368,4 +369,19 @@ export const getContractIdFromTx = (tx?: Transaction | MempoolTransaction) => {
   if (tx.tx_type === 'contract_call') contractId = tx.contract_call.contract_id;
   if (tx.tx_type === 'smart_contract') contractId = tx.smart_contract.contract_id;
   return contractId;
+};
+
+export function initBigNumber(num: string | number | BigNumber) {
+  return BigNumber.isBigNumber(num) ? num : new BigNumber(num);
+}
+
+export const ftDecimals = (value: number | string | BigNumber, decimals: number) => {
+  return initBigNumber(value)
+    .shiftedBy(-decimals)
+    .toNumber()
+    .toLocaleString('en-US', { maximumFractionDigits: decimals });
+};
+
+export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: number) => {
+  return initBigNumber(value).shiftedBy(decimals).toString(10);
 };
--- a/src/components/balances/token-asset-list-item.tsx
+++ b/src/components/balances/token-asset-list-item.tsx
@@ -1,26 +1,43 @@
-import * as React from 'react';
-import { AddressBalanceResponse } from '@stacks/stacks-blockchain-api-types';
+import React, { useEffect, useState } from 'react';
+import { useAtomValue } from 'jotai/utils';
+import { FungibleTokenMetadata } from '@stacks/blockchain-api-client';
 import { Box, color, DynamicColorCircle, Flex, FlexProps, Stack } from '@stacks/ui';
 import { Caption, Text } from '@components/typography';
 import { TxLink } from '@components/links';
-import { getAssetNameParts } from '@common/utils';
+import { ftDecimals, getAssetNameParts, initBigNumber } from '@common/utils';
 import { getTicker } from '@components/tx-events';
+import { apiClientsState } from '@store/api-clients';
 
 interface TokenAssetListItemProps extends FlexProps {
+  amount: string;
   token: string;
-  isLast: boolean;
-  balances: AddressBalanceResponse;
-  type: 'non_fungible_tokens' | 'fungible_tokens';
+  tokenType: 'non_fungible_tokens' | 'fungible_tokens';
 }
-
 export const TokenAssetListItem: React.FC<TokenAssetListItemProps> = ({
+  amount,
   token,
-  balances,
-  type,
-  isLast,
+  tokenType,
 }) => {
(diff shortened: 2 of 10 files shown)


src/common/utils.tsx
```diff
@@ -369,3 +370,20 @@ export const getContractIdFromTx = (tx?: Transaction | MempoolTransaction) => {
   if (tx.tx_type === 'smart_contract') contractId = tx.smart_contract.contract_id;
   return contractId;
 };
+
+export function initBigNumber(num: string | number | BigNumber) {
+  return BigNumber.isBigNumber(num) ? num : new BigNumber(num);
+}
+
+export const ftDecimals = (value: number | string | BigNumber, decimals: number = STX_DECIMALS) => {
```
kyranjamie, line 378: I don't reckon this function should have the default parameter.

src/common/utils.tsx
```diff
@@ -369,3 +370,20 @@ export const getContractIdFromTx = (tx?: Transaction | MempoolTransaction) => {
...
+  const amount = initBigNumber(value);
+  return amount
```
kyranjamie, line 380:
```suggestion
  return initBigNumber(value)
``` 

?

src/components/balances/token-asset-list-item.tsx
```diff
@@ -1,26 +1,44 @@
...
 import { Box, color, DynamicColorCircle, Flex, FlexProps, Stack } from '@stacks/ui';
 import { Caption, Text } from '@components/typography';
 import { TxLink } from '@components/links';
-import { getAssetNameParts } from '@common/utils';
+import { ftDecimals, getAssetNameParts } from '@common/utils';
 import { getTicker } from '@components/tx-events';
+import { apiClientsState } from '@store/api-clients';
 
 interface TokenAssetListItemProps extends FlexProps {
+  balanceOrCount: string;
```
kyranjamie, line 13: Reusing the same prop for different values is best avoided if possible. You could create a conditional, discriminating type here:

```ts
interface TokenAssetListItemFungibleProps extends FlexProps {
  tokenType: 'fungible_tokens';
  balance: string;
  token: string;
}
interface TokenAssetListItemNonFungibleProps extends FlexProps {
  tokenType: 'non_fungible_tokens';
  count: string;
  token: string;
}
type TokenAssetListItemProps = TokenAssetListItemFungibleProps | TokenAssetListItemNonFungibleProps;
```

src/components/balances/token-asset-list-item.tsx
```diff
@@ -1,26 +1,44 @@
...
+        contractId,
+      });
+      setFtMetadata(data);
+    };
+    if (tokenType === 'fungible_tokens') void getFtMetadata();
+  }, []);
+
+  const totalType = tokenType === 'non_fungible_tokens' ? 'count' : 'balance';
+
+  if (new BigNumber(balanceOrCount).isLessThanOrEqualTo(0)) return null;
```
kyranjamie, line 40: Could this ever not be a legit number? Then bignum would throw.

kyranjamie (commented):
Added some suggestions. I don't think a non-DRY solutions will create problems and is okay for now, my feedback is aimed much more towards API decisions.

src/common/utils.tsx
fbwoolf: Yeah, good point, prob not the best way to do it.

src/components/balances/token-asset-list-item.tsx
fbwoolf: Yep, it was originally just `balance` and I renamed it. I'll look at doing your suggestion. I was just trying to keep this simple until someone is really working on the explorer consistently again and can refactor a lot of this code.

src/components/balances/token-asset-list-item.tsx
fbwoolf: I take a closer look at this.

src/common/utils.tsx
fbwoolf: I realized I copied these from the web wallet, so prob should be updated there too at some point.

src/components/balances/token-asset-list-item.tsx
```diff
@@ -1,26 +1,44 @@
...
-  isLast,
+  tokenType,
 }) => {
+  const [ftMetadata, setFtMetadata] = useState<FungibleTokenMetadata | undefined>();
+  const { tokensApi } = useAtomValue(apiClientsState);
   const { address, asset, contract } = getAssetNameParts(token);
-  const key = type === 'non_fungible_tokens' ? 'count' : 'balance';
+
+  useEffect(() => {
+    const getFtMetadata = async () => {
```
beguene, line 28: I think this fetch logic is worth extracting and placing it outside the view.

src/components/post-conditions.tsx
```diff
@@ -239,20 +248,40 @@ const getAddressValue = (condition: PostCondition) => {
 };
 
 const Condition = ({ condition, isLast }: { condition: PostCondition; isLast?: boolean }) => {
+  const [ftMetadata, setFtMetadata] = useState<FungibleTokenMetadata | undefined>();
+  const { tokensApi } = useAtomValue(apiClientsState);
+
+  useEffect(() => {
+    const getFtMetadata = async () => {
```
beguene, line 255: Same here.

src/components/tx-events.tsx
```diff
@@ -252,6 +259,16 @@ const Item: React.FC<{ event: TransactionEvent; isLast?: boolean }> = ({ event,
       ? 'Non-fungible token'
       : undefined;
 
+  useEffect(() => {
+    const getFtMetadata = async () => {
```
beguene, line 263: Same here.

src/components/balances/token-asset-list-item.tsx
beguene: You could maybe use the word 'amount' instead of balanceOrCount (that's what we do in the wallet) and make this view pure. The calculation logic can be done outside and then only display the values passed to this view component (like AssetItem in the wallet)

src/components/balances/token-asset-list-item.tsx
fbwoolf: I had thought abt it but have to send in the tokensApi as an arg which seemed a bit strange but can do.

src/components/balances/token-asset-list-item.tsx
fbwoolf: Ok, thx for the suggestion, I'll check out the web wallet implementation.

Issue #561, opened 2021-10-06
Fix "View schedule" for addresses with unlocking tokens


andresgalante:
This is a regression, the "view schedule" button isn't working here:

https://explorer.stacks.co/address/SP2HR7KFZA9XF6YK33M83JVKB7FZA5G5XAQJ24T31?utm_source=stacks-wallet&chain=mainnet

Inspecting the code it looks like it's not even a link

![](https://cdn.discordapp.com/attachments/803293319328170024/895387632660545556/Screen_Shot_2021-10-06_at_16.10.16.png)

fbwoolf:
@andresgalante the code is here and is called by an `onClick` function `handleOpenUnlockingScheduleModal`: https://github.com/blockstack/explorer/blob/91578e08528b06b1b52bb2ce9eac18c555b01278/src/components/balances/stx-balance-card.tsx#L160

andresgalante:
Is this issue Done?

fbwoolf:
This is done but it hasn't been released yet.

fbwoolf:
I will close the issue. @He1DAr if you link the issue in your PR when you merge it the issue will automatically close.

PR #614, opened 2022-01-07
fix: View schedule for addresses with unlocking tokens

He1DAr:
We currently use `withInitialQueries` from `jotai-query-toolkit`, this HOC [wraps the page into a new Jotai provider](https://github.com/fungible-systems/jotai-query-toolkit/blob/3cdba6c65d09c615b1851a2f9c02db3273c60848/src/nextjs/intial-queries-wrapper.ts#L30-L34).
This creates an isolated state for each wrapped page, which prevents it from updating the state outside page scope (in this case, modal state).
A **temporary** fix could be to move this modal to the pages where it's being used. We probably should aim towards eliminating `jotai-query-toolkit` in the near future.
#561

### DIFF
--- a/src/components/modals.tsx
+++ b/src/components/modals.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { SearchModal } from '@components/modals/search';
 import { AddNetworkModal } from '@components/modals/add-network';
-import { UnlockingScheduleModal } from '@components/modals/unlocking-schedule';
 import { SafeSuspense } from '@components/ssr-safe-suspense';
 
 export const Modals: React.FC = () => {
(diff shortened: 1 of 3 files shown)


fbwoolf (approved):
Nice work @He1DAr! I agree, we should move toward removing the toolkit, but for now this looks like a good temp fix until we can spend more time refactoring.

fbwoolf:
@He1DAr if you run `lint:prettier:fix` it should fix your lint errors. And, we have a standard for commit messages. I believe yours is failing on the capital `V` in the subject ...if you amend the commit message it should pass that check.

`subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]`

kyranjamie (approved):
Good stuff Heidar.

You'll need to set up your IDE auto-formatter so it uses our prettier config

Issue #622, opened 2022-01-21
Show nonces on (pending) transactions viewing an address in the explorer


314159265359879:
**Is your feature request related to a problem? Please describe.**
I have been troubleshooting a lot to help new users with trouble getting their transactions processed. On rare occasions transactions get send with the wrong nonce or a gap has dropped due to a transaction dropping from the mempool. 
At the moment you have to click the individual transactions that have been processed and that are pending to see the nonce they were send with.

**Describe the solution you'd like**
a crude mockup:

![image](https://user-images.githubusercontent.com/33360391/150611909-0825122c-37be-4490-b172-feded7fa6c04.png)

By displaying it here you have all the nonces in sight and you can easily see nonces are as expected. The nonce can be displayed small and/or obscured (light grey on black/white) as most users will not need it. It does not need a prominent place.

**Describe alternatives you've considered**
n.a.

Hero-Gamer:
Agree having Nonce in the Explore tx history overview page will help in support trouble shooting
I'd also suggest displaying the fee. 

And I would further suggest both nonce and fee be added to Hiro wallet Activity itself, so all tx (or the key ones) have Nonce and fee used. This will support quicker diagnose and action taken to fix clogged wallets. 

Look forward to hearing your thoughts.

PR #624, opened 2022-01-22
feat: add nonce indicator on pending transactions when viewing an address (#622)

linden:
Adds a indicator next to pending transactions, reasons detailed here (#622).

### DIFF
--- a/src/components/transaction-item.tsx
+++ b/src/components/transaction-item.tsx
@@ -158,6 +158,17 @@ const Timestamp: React.FC<BoxProps & { tx: Transaction | MempoolTransaction }> =
   }
 );
 
+const Nonce: React.FC<BoxProps & { nonce: Number, isHovered?: boolean }> = React.memo(
+  props => (
+      <>
+        {'·'}
+        <Caption as="span" text-align="right" ml="6px">
+         {props.nonce}
+        </Caption>
+      </>
+    )
+);
+
 const LargeVersion = React.memo(
   ({
     tx,
@@ -205,9 +216,14 @@ const LargeVersion = React.memo(
                 {didFail && 'Failed'}
               </Caption>
               {'·'}
-              <Caption mt="1px" ml="6px">
+              <Caption mt="1px" ml="6px" mr={isPending && "6px"}>
                 {truncateMiddle(tx.tx_id, 4)}
               </Caption>
+
+              { isPending &&
+                <Nonce ml="6px" nonce={tx.nonce} isHovered={isHovered}>
+                </Nonce>
+              }
             </Flex>
           </Stack>
         ) : null}
--- a/src/components/transaction-item.tsx
+++ b/src/components/transaction-item.tsx
@@ -165,7 +165,7 @@ const Nonce: React.FC<BoxProps & { nonce: Number, isHovered?: boolean }> = React
         {'·'}
         <Tooltip label="Nonce">
           <Caption as="span" text-align="right" ml="6px" style={{zIndex:1}}>
-            {props.nonce}
+            {props.nonce + "n"}
           </Caption>
         </Tooltip>
       </>
--- a/src/components/transaction-item.tsx
+++ b/src/components/transaction-item.tsx
@@ -163,7 +163,7 @@ const Nonce: React.FC<BoxProps & { nonce: Number, isHovered?: boolean }> = React
   props => (
       <>
         {'·'}
-        <Tooltip label="nonce">
+        <Tooltip label="Nonce">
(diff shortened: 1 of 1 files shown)


linden:
I've got 3 options for how this could be displayed:

A:
image
Simple, although looks awkward when compared to non-pending transactions.

B: 
image
Easy to associate with a pending transaction.

C:
image
Blends in, would need to add some sort of hover-over description.

Any feedback on what you all think fits the best?

kyranjamie:
Great stuff, thanks for the contribution. Def C with a tooltip. Maybe the number styleized as e.g. `17n`? 🤔

314159265359879:
Thanks for picking this up @linden here are my thoughts:

A is most descriptive, but I like C most because it least noticeable and I think that is fine in most cases. I think it is clever to put in a line we were already using as to not make the increase the height required per item shown. 

What about C including the word _nonce_? 
_Pending - 0xa31d...f886 - nonce: 14_

314159265359879:
What Kyran says even better `14n`, the hover will do the rest.

fbwoolf:
> Great stuff, thanks for the contribution. Def C with a tooltip. Maybe the number styleized as e.g. `17n`? 🤔

I like this idea bc on the home page there is limited space across. I don't think spelling out `nonce` is necessary? I would say `C` is best and add the `n`

linden:
Thanks for the feedback! Here's what it looks like with the tooltip and the `n`.

https://user-images.githubusercontent.com/70739041/150647564-3f206042-2354-4dfe-a658-b3786e21c003.mov

314159265359879 (approved):
LGTM

Hero-Gamer:
> > Great stuff, thanks for the contribution. Def C with a tooltip. Maybe the number styleized as e.g. `17n`? 🤔
> 
> I like this idea bc on the home page there is limited space across. I don't think spelling out `nonce` is necessary? I would say `C` is best and add the `n`

I'll speak from users perspective if I may?

as much as I like "n"
I think spelled out "Nonce" is easier for the less advanced users, which there are a lot across the ecosystem. One less thing for us having to explain. There should be plenty of space on the page (majority uses desktop to access Explorer)

I'd also like to suggest add "Fees" to the Explorer. Both Explorer and Hiro wallet if possible. 
Sorry!

I've made the comment on separate issues: 
https://github.com/hirosystems/explorer/issues/622
https://github.com/hirosystems/stacks-wallet-web/issues/2129

landitus (approved):
Even though I see good points for making the nonce more discoverable for new users, I'm thinking it's best to keep the line items as simple as possible given they repeat multiple times.  Right now, each information doesn't have a label, which is a good decision to avoid repeating text and crowding the interface. 

Having the tooltip as shown [here](https://github.com/hirosystems/explorer/pull/624#issuecomment-1019306403) should do the trick now.

For adding more information, such as fees, it will be more challenging (since there are no labels to explain what each piece of information means). I think then it may be time to review the line item's design a bit further.

src/components/transaction-item.tsx
```diff
@@ -205,9 +219,14 @@ const LargeVersion = React.memo(
...
               </Caption>
               {'·'}
-              <Caption mt="1px" ml="6px">
+              <Caption mt="1px" ml="6px" mr={isPending && "6px"}>
                 {truncateMiddle(tx.tx_id, 4)}
               </Caption>
+
+              { isPending &&
+                <Nonce ml="6px" nonce={tx.nonce}>
+                </Nonce>
```
He1DAr, line 228: nit:
```suggestion
                <Nonce ml="6px" nonce={tx.nonce}/>
```

src/components/transaction-item.tsx
```diff
@@ -158,6 +159,19 @@ const Timestamp: React.FC<BoxProps & { tx: Transaction | MempoolTransaction }> =
   }
 );
 
+const Nonce: React.FC<BoxProps & { nonce: Number, isHovered?: boolean }> = React.memo(
```
He1DAr, line 162: Does this component only need `nonce`? 
```suggestion
const Nonce: React.FC<{ nonce: Number }> = React.memo(
```

He1DAr (commented):
LGTM! left a couple of small comments. Could you resolve CI issues?

src/components/transaction-item.tsx
linden: Probably, I'm new to React so I was really winging this PR.

linden:
Looks ready to me, anything else needed @He1DAr?

314159265359879:
This is brilliant, I love it! 
Makes it so much easier to troubleshoot nonce issues. 
image

PR #634, opened 2022-02-04
refactor: remove modal recoil state and introduce redux

### LINKED ISSUE #635: refactor: remove modal recoil state and introduce redux
### QA:
All modals should be working exactly like the current state.

He1DAr:
- Introducing Redux
- Refactoring modal recoil state
closes #635


fbwoolf (approved):
Looks like a nice way to start this process. 👍

src/components/modals/add-network.tsx
```diff
@@ -2,18 +2,20 @@ import React from 'react';
 import { Box, color, ControlledModal, Flex, IconButton, Stack } from '@stacks/ui';
 import { IconX } from '@tabler/icons';
 import { Link, Text, Title } from '@components/typography';
-import { useModal } from '@common/hooks/use-modal';
+import { closeModal, getOpenedModal } from '@common/state/slices/modal';
```
beguene, line 5: @He1DAr I thought we would prefer organising folders by features from now on.

beguene (approved):

src/components/modals/add-network.tsx
kyranjamie: +1 on this

Wouldn't `common/state/modal/modal.slice.ts` be more organising by feature?

src/common/state/slices/modal.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+    closeModal: state => {
+      state.openedModal = undefined;
+    },
+  },
+});
+
+export const { openModal, closeModal } = slice.actions;
+
+export const getOpenedModal = () => useAppSelector(state => state.modal.openedModal);
```
kyranjamie, line 28: Should we figure out a naming convention for selectors? I've been using `select` Prefixed naming.

package.json
```diff
@@ -30,6 +30,7 @@
     "@react-aria/interactions": "3.5.0",
     "@react-spectrum/utils": "3.6.1",
     "@react-stately/toggle": "3.2.2",
+    "@reduxjs/toolkit": "^1.7.2",
```
kyranjamie, line 33: Try and use the `-E` flag when installing packages, so it pins to an exact version. This was a recommendation of a security firm (we have a CI job to lint this)

kyranjamie (approved):
Looks good @He1DAr, added a few points to discuss.

Modal makes sense to start with, as this was already handled with state, though I think it's best practice that modals have corresponding routes, (rather than state)

src/components/modals/add-network.tsx
He1DAr: I was under the impression modal is a global state, but looking now we can def colocate it with modal 👍

He1DAr:
> Looks good @He1DAr, added a few points to discuss.
> 
> Modal makes sense to start with, as this was already handled with state, though I think it's best practice that modals have corresponding routes, (rather than state)

I like the modal routes, what are your thoughts on keeping the same functionality with this refactor and addressing modal routes as a separate feature?

src/common/state/slices/modal.ts
He1DAr: I'm good with both, I used `get` as it's more framework-agnostic (not associated with reselect) but I think both work well.

kyranjamie:
> I like the modal routes, what are your thoughts on keeping the same functionality with this refactor and addressing modal routes as a separate feature?

Agreed, this isn't necessary for now.

src/common/state/slices/modal.ts
kyranjamie: @beguene @fbwoolf for input

I don't really care either way, but let's just pick one.

fbwoolf: I would just go with `select` if it is already implemented in the wallet changes. It makes sense to me as prefix.

beguene: I prefer ```select```.

Issue #625, opened 2022-01-23
Explorer displays MillionSTX as ThousandSTX


sleepiJoe:
**Describe the bug**

Address: SP1P72Z3704VMT3DMHPP2CB8TGQWGDBHD3RPR9GZS

**To Reproduce**
1. Go to [this wallet](https://explorer.stacks.co/address/SP1P72Z3704VMT3DMHPP2CB8TGQWGDBHD3RPR9GZS?chain=mainnet)
2. See error

**Expected behavior**
Normally it should show as 8.015.897,57 STX.

**Screenshots**
![image](https://user-images.githubusercontent.com/94863769/150688563-3d3b1120-3796-4bb8-a16d-f519f844eacc.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome
 - Version v97

**Additional context**
NA

fbwoolf:
This is what I see:

<img width="1250" alt="Screen Shot 2022-01-23 at 11 47 56 AM" src="https://user-images.githubusercontent.com/6493321/150691214-d3231088-a878-440e-a1b7-9b6c585dcda7.png">

sleepiJoe:
> This is what I see:

Which browser do you use? Because in Firefox it's displayed correctly, then in Chrome not.

fbwoolf:
That screenshot is taken from Chrome.

sleepiJoe:
@fbwoolf 

html:
`<div class="css-1bw9tv8"><span style="user-select:all" class="css-fyja9i"><span class="css-18td2fn">8</span><span class="css-1ydounk">.<!-- -->037</span><span class="css-12g23ss">STX</span></span><span style="user-select:none" class="css-ks4f0t">Total balance</span></div>`

Some js loaded from explorer.stacks.co breaks the originally delivered html code into the above example.

LunarPunks:
> **Describe the bug** A clear and concise description of what the bug is.
> 
> Address: SP1P72Z3704VMT3DMHPP2CB8TGQWGDBHD3RPR9GZS
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
> 1. Go to [this wallet](https://explorer.stacks.co/address/SP1P72Z3704VMT3DMHPP2CB8TGQWGDBHD3RPR9GZS?chain=mainnet)
> 2. See error
> 
> **Expected behavior** Normally it should show as 8.015.897,57 STX.
> 
[...]

PR #636, opened 2022-02-08
fix: locale decimal separator issue

chxash:
Fix for https://github.com/hirosystems/explorer/issues/625
Fix multiple issues at once related to localization.

decimalgerman

### DIFF
--- a/src/common/utils.tsx
+++ b/src/common/utils.tsx
@@ -385,3 +385,7 @@ export const ftDecimals = (value: number | string | BigNumber, decimals: number)
 export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: number) => {
   return initBigNumber(value).shiftedBy(decimals).toString(10);
 };
+
+export function getLocaleDecimalSeparator() {
+  return (1.1).toLocaleString().substring(1, 2);
+}
--- a/src/common/utils.tsx
+++ b/src/common/utils.tsx
@@ -387,5 +387,5 @@ export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: 
 };
 
 export function getLocaleDecimalSeparator() {
-  return (1.1).toLocaleString().substring(1, 2);
+  return Intl.NumberFormat().formatToParts(1.1).find(part => part.type === 'decimal')?.value
 }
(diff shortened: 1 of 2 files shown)


src/common/utils.tsx
```diff
@@ -385,3 +385,7 @@ export const ftDecimals = (value: number | string | BigNumber, decimals: number)
 export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: number) => {
   return initBigNumber(value).shiftedBy(decimals).toString(10);
 };
+
+export function getLocaleDecimalSeparator() {
+  return (1.1).toLocaleString().substring(1, 2);
```
kyranjamie, line 390: Could you use the `Intl` here? I'm not sure myself, but might be appropriate https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

kyranjamie (approved):

src/common/utils.tsx
chxash: done ! same thing with Intl

src/common/utils.tsx
```diff
@@ -385,3 +385,7 @@ export const ftDecimals = (value: number | string | BigNumber, decimals: number)
 export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: number) => {
   return initBigNumber(value).shiftedBy(decimals).toString(10);
 };
+
+export function getLocaleDecimalSeparator() {
+  return Intl.NumberFormat().formatToParts(1.1).find(part => part.type === 'decimal')?.value
```
zone117x, line 390: Is this function guaranteed to _only_ run/render on the client? My understanding is that all or any of the explorer pages can be rendered on the server via nextjs, and in those cases, this will use whatever the default nodejs server locale is (something like en-US iirc).

/cc @kyranjamie @He1DAr

chxash: @zone117x  Indeed, it seems this function is called on server side as well, before being called on client side.
This has no effect since the correct locale is used on client side. I have tested multiple languages and it works as expected.
We could return a default value or return an error when called from server side if you prefer.

andresgalante:
@He1DAr Would you mind ushering this PR to prod please?

fbwoolf (approved):

fbwoolf:
Thank you for the contribution @chxash. I tested this and will merge it into main now. It will likely get released to production with some changes we are currently working on to the search bar this week.

PR #660, opened 2022-03-16
fix: move suspense

fbwoolf:
The application error in the linked issue is hard to reproduce with a running local env, so debugging is tricky on this. The error is telling us to move the Suspense boundary higher in the component tree. I believe this was (perhaps) introduced with the changes in this PR? https://github.com/hirosystems/explorer/pull/614

@He1DAr any thoughts on this solution to try? I can't say for sure it fixes the error bc I can't reproduce it locally. 

Screen Shot 2022-03-16 at 9 19 05 AM

### DIFF
--- a/src/components/modals.tsx
+++ b/src/components/modals.tsx
@@ -1,15 +1,12 @@
 import React from 'react';
 import { SearchModal } from '@components/modals/search';
 import { AddNetworkModal } from '@components/modals/add-network';
-import { SafeSuspense } from '@components/ssr-safe-suspense';
 
 export const Modals: React.FC = () => {
   return (
     <>
-      <SafeSuspense fallback={<></>}>
-        <SearchModal />
-        <AddNetworkModal />
-      </SafeSuspense>
+      <SearchModal />
+      <AddNetworkModal />
     </>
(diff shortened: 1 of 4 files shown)


He1DAr (approved):

He1DAr:
Thank you for looking into this. It could happen because we have multiple elements inside a suspense block. Couple of issues we need to investigate:
1) Why this error is not showing in Sentry?
2) Why source map is not working on this instance?

fbwoolf:
> 
> 1. Why this error is not showing in Sentry?
> 2. Why source map is not working on this instance?

I will look into these before we merge this.

fbwoolf:
I am going to deploy this to staging to test.

fbwoolf:
@He1DAr

I think we need to add this to get the error in Sentry:
https://docs.sentry.io/platforms/javascript/guides/react/components/errorboundary/

And, I think we need to add this to get source maps in production:
https://nextjs.org/docs/advanced-features/source-maps

fbwoolf:
Created a new issue for these: https://github.com/hirosystems/explorer/issues/669

PR #673, opened 2022-03-28
refactor: search bar

He1DAr:


src/features/search/useSearch.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { searchApi } = useAtomValue(apiClientsState);
+  const id = searchTerm ? debouncedSearchTerm : '';
+  const query = useQuery(
+    ['search', id],
+    async () => {
+      try {
+        const result = await searchApi.searchById({ id });
+        return result as FoundResult;
+      } catch (e) {
+        try {
```
beguene, line 21: I think this error handling logic could be extracted as we might use it more and more later.

beguene (approved):
Nice refactor + cleanup ! 👍

src/common/hooks/useDebounce.ts
```diff
@@ -0,0 +1,17 @@
+import { useState, useEffect } from 'react';
```
kyranjamie, line 1: Filename should be `use-debounce.ts` per naming convention.

src/features/search/useSearch.ts
kyranjamie: +1, I'd go further and suggest two hooks. This one, a higher-level abstraction used in the app. Plus one that handles just the `useSearchQuery` that returns all the react query options. [wallet example](https://github.com/hirosystems/stacks-wallet-web/tree/dev/src/app/query/mempool)

src/features/search/searchSlice.ts
```diff
@@ -0,0 +1,43 @@
+import { createSelector, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { RootState } from '@common/state/store';
+
+export interface State {
```
kyranjamie, line 4: Given this is exported and used externally, it'd be better to give it a specific name
```suggestion
export interface SearchState {
```

src/features/search/searchSlice.ts
```diff
@@ -0,0 +1,43 @@
...
+  searchTerm: string;
+  isFocused: boolean;
+}
+
+const initialState: State = {
+  searchTerm: '',
+  isFocused: false,
+};
+
+export const slice = createSlice({
```
kyranjamie, line 14:
```suggestion
export const searchSlice = createSlice({
```

src/features/search/searchSlice.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const selectSearch = (state: RootState) => state.search;
+
+export const selectSearchTerm = createSelector([selectSearch], search => search.searchTerm);
+export const selectIsSearchFieldFocused = createSelector(
+  [selectSearch],
+  search => search.isFocused
+);
+
+export default slice.reducer;
```
kyranjamie, line 43: Please let's avoid default exports. They make searching through a codebase extremely hard.

src/features/search/useSearch.ts
kyranjamie: Also file name introduces inconsistency

src/features/search/searchSlice.ts
```diff
@@ -0,0 +1,43 @@
...
+  isFocused: false,
+};
+
+export const slice = createSlice({
+  name: 'search',
+  initialState,
+  reducers: {
+    setSearchTerm: (state, action: PayloadAction<string>) => {
+      state.searchTerm = action.payload;
+    },
```
kyranjamie, line 20: What's the requirement that the search term is persisted in the store? Do we need to later?

Generally form values should be kept out of redux imo, [See Dan Abromov comment](https://github.com/reduxjs/redux/issues/1287#issuecomment-175351978)

kyranjamie (changes requested):
Good work @He1DAr, this is a good refactor

Consider comments on use of redux optional, though rejecting because introduction of file naming in consistencies, and use of default exports. 

[Here's one article I've picked at random making the general arguments against default exports](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad). Being able to name a variable one thing, and import it named  something completely different makes navigating/searching codebases _way_ harder. We've avoided this in all our codebases. There are some cases where renaming is useful, this can be done with named exports using the `as` keyword.

src/features/search/searchSlice.ts
kyranjamie: https://goshacmd.com/should-i-put-form-state-into-redux/

fbwoolf: @kyranjamie do you have a link to the diagram on when to use what state mgmt? Was that in notion?

kyranjamie: The one I made is in this document https://www.notion.so/hiropbc/State-management-next-steps-9027c10831f54cd8bf40d4566281b0f1

src/features/search/useSearch.ts
He1DAr: > +1, I'd go further and suggest two hooks. This one, a higher-level abstraction used in the app. Plus one that handles just the `useSearchQuery` that returns all the react query options. [wallet example](https://github.com/hirosystems/stacks-wallet-web/tree/dev/src/app/query/mempool)

@kyranjamie what's the motive behind moving the query to a different file if it's only used here anyway (and the file size is small)?

src/components/modals.tsx
```diff
@@ -1,12 +1,6 @@
...
 import { AddNetworkModal } from '@components/modals/add-network';
 
 export const Modals: React.FC = () => {
-  return (
-    <>
-      <SearchModal />
-      <AddNetworkModal />
-    </>
-  );
+  return <AddNetworkModal />;
```
fbwoolf, line 5: Do you think we even need to keep this file? Maybe just relocate the one modal component or do we intend to add to this pattern in the future?

src/features/search/useSearch.ts
kyranjamie: I'm not so concerned about it being a different file/same file. Motivation is more code reuse/SRP/loose coupling of the hook.

This hook:
- Interfaces with the store
- fetches data from the api

These could be thought of as distinct functionalities. Consider we want to reuse the search query hook elsewhere, and that it shouldn't touch an atom. We can't without refactoring.

Better that `useSearch` composes the functionality of smaller hooks, than putting everything in one big one.

src/components/transaction-item.tsx
```diff
@@ -121,14 +121,9 @@ const AddressArea = React.memo(
   }
 );
 
-const TxItemTitleArea = React.memo(({ title, isHovered, tx, principal }: any) => (
+const TxItemTitleArea = React.memo(({ title, tx, principal }: any) => (
```
fbwoolf, line 124: @kyranjamie what was the change you made in the wallet where we don't need to import `React` anymore? I think that would help clean up the explorer too.

kyranjamie: https://github.com/hirosystems/stacks-wallet-web/pull/2091

src/features/search/searchSlice.ts
He1DAr: @kyranjamie I generally agree with using react state for form inputs, this one is a bit different though as I use the same state for both search bars (homepage and header) and result wrappers. If we use react state, we'll need to call useSearch hook in the parent component and pass props down to search bar and result wrapper (which is not the best IMO) + write it twice in both search bars. Or maybe there's a better idea?

src/components/modals.tsx
fbwoolf: Not necessary if you think we will add to it later. 👍

src/components/transaction-item.tsx
fbwoolf: @He1DAr I am going to make an issue to do this in the explorer.

kyranjamie (approved):
LGTM

fbwoolf:
@He1DAr is this ready to release?

He1DAr:
> @He1DAr is this ready to release?

waiting to hear back from QA before merging this

PR #689, opened 2022-04-22
ci: docker image digests

CharlieC3:
Leverages a new feature in our deployment action to specify Docker image digests instead of the general tag on CI deployments.
This helps resolve the scenario where two commits are pushed to a branch, and the first commit gets deployed, but the second one results in a no-op deployment since the tag didn't change.

Resolves https://github.com/hirosystems/devops/issues/927

[Working](https://github.com/hirosystems/explorer/actions/runs/2204240909)


He1DAr (approved):
LGTM, thank you for fixing this! 🎉

PR #694, opened 2022-04-26
feat: add GA and update footer links

He1DAr:


andresgalante:
Maybe @tprebz can take a look at this one and approve it?

PR #706, opened 2022-05-10
fix: refactor transaction page

He1DAr:
Applying same refactor to transaction and address page


fbwoolf:
@He1DAr can you provide any context in the description about the changes being made here? Link any issues?

PR #721, opened 2022-05-18
fix: network state naming and default networks persistence

He1DAr:
- remove default networks from persistent state
- fix naming inconsistencies
- fix RootState type


kyranjamie (approved):
Nice tidy up

PR #723, opened 2022-05-19
fix: clarinet instance API url

He1DAr:
Fix clarinet instance API url


src/common/utils.tsx
```diff
@@ -412,3 +417,8 @@ export function getContractId(
     ? transaction.contract_call.contract_id
     : undefined;
 }
+
+export const isClarinetInstance = (url: string) => url.includes('.devnet:3999');
```
kyranjamie, line 421: Is there scope for abuse in this check? 🤔  Would it be better/safer to pass url to `new URL()` and check host?

kyranjamie (approved):

PR #725, opened 2022-05-24
feat: skip prefetch for devnet

He1DAr:
Exploring a fix to devnet (localhost) server-side data fetching. Basically skipping prefetch, which will cause the client to do the requests instead of the server.

### DIFF
--- a/src/pages/address/[principal].tsx
+++ b/src/pages/address/[principal].tsx
@@ -16,11 +16,12 @@ import { AddressNotFound } from '@components/address-not-found';
 import { UnlockingScheduleModal } from '@components/modals/unlocking-schedule';
 import { ServerResponse } from 'http';
 import { QueryClient, useQuery } from 'react-query';
-import { wrapper } from '@common/state/store';
+import { store, wrapper } from '@common/state/store';
 import { dehydrate } from 'react-query/hydration';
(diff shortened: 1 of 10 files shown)


fbwoolf (approved):
Looks like a good solution.

src/pages/address/[principal].tsx
```diff
@@ -111,52 +112,54 @@ const AddressPage: NextPage<any> = ({ error }) => {
...
-      ),
-      queryClient.prefetchQuery(
-        addressQK(AddressQueryKeys.accountBalance, addressPageQuery),
-        queries.fetchAccountBalance(addressPageQuery),
-        prefetchOptions
-      ),
-    ]);
-  } catch (err) {
-    res.statusCode = err.status;
+  if (networkUrl) {
```
Comment (beguene) [inline src/pages/address/[principal].tsx:120]: I often prefer using guard clause to keep the main code logic flat. I think it's a bit easier to read and prevents too deeply nested logic.
```javascript
if (!networkUrl) return
const prefetchOptions = { staleTime: 5000 };
const queries = getAddressQueries(networkUrl);
try {
...
```
Same comment below

beguene (approved):
Nice one @He1DAr 👍, it seems like a good and simple solution and maybe a good template to follow for those kind of problems (which will definitively occurs in the future). Besides it will work straight out of the box for the developer when working locally.

PR #731, opened 2022-05-27
Fix/add prometheus metrics/i329

beguene:


kyranjamie (approved):

beguene:
@CharlieC3 Is there a way to open the port 9153 in the host so we can test it on staging ? (it should also be open in production)

He1DAr:
We probably should look into using something that doesn't require a custom next.js server. Using a custom server is discouraged and can add extra complexity.
https://nextjs.org/docs/advanced-features/custom-server
image

beguene:
@He1DAr I would also prefer not using a custom server. Let's see which infra we are targeting and come back to this later.
We might not even need prometheus.

andresgalante:
@CharlieC3 it doesn't seem like we'll be able to introduce Prometheus to the explorer. Is this ok or would you like us to meet and brainstorm options to get it done?

CharlieC3:
> It doesn't seem like we'll be able to introduce Prometheus to the explorer. 

@andresgalante Is this because of the concerns around it being implemented as a custom server? If so, are we sure the optimizations lost when using a prometheus custom server would affect the whole service's performance? Additionally, is it a matter of a couple milliseconds or something more significant?

> Is this ok or would you like us to meet and brainstorm options to get it done?

Having app metrics for the Explorer would be most valuable to the UX team, so I feel your team should decide if it's ok or not.
DevOps doesn't necessarily _need_ to have them implemented for the Explorer, but we highly suggest it as it may help UX discover bugs and measure changes over time like it has for Splat with the API. And as the Explorer gets more complex, it may be crucial to debug issues and restore availability for any UX team members that are on call.

PR #737, opened 2022-05-29
fix: pending contract deployment

### LINKED ISSUE #736: contract deployment pending tx not displaying
when deploying smart contract and visiting tx while pending, it displays a blank page.

He1DAr:
closes #736


kyranjamie (approved):
This now returns `undefined` rather than `null`. How does this fix it, exactly?

He1DAr:
> This now returns `undefined` rather than `null`. How does this fix it, exactly?

I removed the return statement if `!contract`, so the page will render normally except for the source code section.

PR #745, opened 2022-06-01
Devops-940

deantchi:
attempting to streamline GH Action build by running the `Semantic Release` before the build an release tasks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,6 @@ on:
   push:
     branches:
       - '**'
-    tags:
-      - 'v*'
     paths-ignore:
       - '**/CHANGELOG.md'
       - '**/package.json'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,6 +40,16 @@ jobs:
...
+        uses: cycjimmy/semantic-release-action@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          SEMANTIC_RELEASE_PACKAGE: ${{ github.event.repository.name }}
+        with:
+          extra_plugins: |
+            @semantic-release/changelog
+            @semantic-release/git
+
       - name: Docker meta
```
CharlieC3, line 53: After this point, the GH tag should be created. We previously discussed switching the local checkout to that git tag before continuing, but I remembered a cleaner way to generate the correct Docker tags.

If you update [these lines](https://github.com/hirosystems/explorer/blob/a2f2b833c3af580147683a5a05e4f81900b29ceb/.github/workflows/ci.yml#L61-L62) to something like [this](https://github.com/hirosystems/stacks-blockchain-api/blob/master/.github/workflows/ci.yml#L529-L530), it should generate the correct tags, which then get fed into the `Build/Tag/Push Image` step.

CharlieC3 (changes requested):
Looking good, just one change is needed to generate the correct Docker tags.

PR #747, opened 2022-06-02
Devops 940

deantchi:
attempting to streamline CI build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,12 @@ jobs:
             @semantic-release/changelog
             @semantic-release/git
 
+      - name: Checkout tag
+        if: ${{ steps.semantic.outputs.new_release_version }} > ${{ steps.semantic.outputs.last_release_version }}
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
+
       - name: Docker meta
         id: docker_meta
         uses: docker/metadata-action@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,6 +40,23 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          SEMANTIC_RELEASE_PACKAGE: ${{ github.event.repository.name }}
+        with:
+          extra_plugins: |
+            @semantic-release/changelog
+            @semantic-release/git
+
+      - name: Checkout tag
+        if: ${{ steps.semantic.outputs.new_release_version }} > ${{ steps.semantic.outputs.last_release_version }}
```
CharlieC3, line 54: Not sure this kind of comparison will work, since it's comparing strings and not numbers (`1.0.1 > 1.0.0 vs 101 > 100)`. I would suggest changing this:
```suggestion
        if: steps.semantic.outputs.new_release_version != ''
```

If there is no new release created, then `steps.semantic.outputs.new_release_version` will be empty and this step will be skipped. Also, the `${{ }}` brackets are optional for the `if` property in an action.

CharlieC3 (changes requested):

.github/workflows/ci.yml
deantchi: changes made
also had to add
```id: semantic```
to enable outputs from the semantic release task

CharlieC3 (approved):
LGTM! :shipit:

PR #748, opened 2022-06-06
ci: auto-approve stg based on new sem ver

deantchi:
- expose new out put in build job
- auto-approve deploy into stg based on new sem version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
     outputs:
       docker_image_digest: ${{ steps.docker_push.outputs.digest }}
       version: ${{ steps.docker_meta.outputs.version }}
+      sem_version_bool: ${{ steps.semantic.outputs.new_release_version }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
     outputs:
       docker_image_digest: ${{ steps.docker_push.outputs.digest }}
       version: ${{ steps.docker_meta.outputs.version }}
+      sem_version_bool: ${{ steps.semantic.outputs.new_release_version }}
```
CharlieC3, line 37: The `new_release_version` var in this action is the new version created. `new_release_published` is a boolean though, maybe you mean this?
https://github.com/marketplace/actions/action-for-semantic-release#outputs

.github/workflows/ci.yml
```diff
@@ -122,7 +123,7 @@ jobs:
 
   auto-approve-staging:
     runs-on: ubuntu-latest
-    if: startsWith(github.ref, 'refs/tags/v')
+    if: ${{ needs.semantic.outputs.sem_version_bool == 'true' }}
```
CharlieC3, line 126: Since `auto-approve-staging` is a different job, it needs to reference the originating job from where the output var. See here for an example:
https://github.com/hirosystems/explorer/blob/72b71f3f0fc6272d6f0ab4a180c738cbde1b58f9/.github/workflows/ci.yml#L120

Also, brackets aren't needed for the `if` property.

CharlieC3 (changes requested):

CharlieC3:
@deantchi I think the condition for the `deploy-prod` job should be updated to be the same as well.

CharlieC3 (approved):

PR #760, opened 2022-06-23
fix: v2 endpoint queried too often

### LINKED ISSUE #586: Prevent /v2 from being queried too often on accounts page
**Is your feature request related to a problem? Please describe.**
On the explorer's `address` page ([example](https://explorer.stacks.co/address/SM2MARAVW6BEJCD13YV2RHGYHQWT7TDDNMNRB1MVT?chain=mainnet)), the `v2` endpoint `https://stacks-node-api.stacks.co/v2/accounts/XXXXXXXXX?proof=0&unanchored=true` is fetched roughly every 10 seconds. With so many people using the explorer, this could create RPC congestion for stacks-nodes.

**Describe the solution you'd like**
I'm not sure what data is being used from `https://stacks-node-api.stacks.co/v2/accounts/XXXXXXXXX?proof=0&unanchored=true` every 10 seconds, but can it be obtained from a `v1` endpoint instead? `v1` is much more performant, and will help stacks-nodes propagate transactions faster with less RPC congestion.

Additionally, I feel a 10 second refresh might be too frequent. If this refresh interval could be changed via environment variable, it would allow those hosting it to configure a refresh interval that balances speed and traffic.

**Side question** - Looks like a few other `v1` endpoints are being called at regular intervals in the explorer as well. Can the explorer  start leveraging websockets? The API now supports websockets, which would be more performant than standard HTTP requests, and data would be transmitted as soon as it's received.

cc @zone117x @rafaelcr

beguene:
closes #586

replace the v2 endpoint with a v1 /nonce fetch

### DIFF
--- a/src/common/api/account.ts
+++ b/src/common/api/account.ts
@@ -1,0 +1,6 @@
+import { fetchFromSidecar } from './fetch';
+
+export const fetchNonce = (apiServer: string) => async (principal: string) => {
+  const res = await fetchFromSidecar(apiServer)(`/address/${principal}/nonces`);
+  return res.json();
+};
--- a/src/common/api/blocks.ts
+++ b/src/common/api/blocks.ts
@@ -21,6 +21,7 @@ export interface FetchBlocksListResponse {
  *
  * @param {string} apiServer - the current apiServer
  */
+// TOREMOVE unused
 export const fetchBlock =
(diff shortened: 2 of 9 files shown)


beguene:
Closing as it's already done in a previous PR.

PR #811, opened 2022-08-08
feat: BTC and STX block height links

He1DAr:
Create a new component `<BtcStxBlockLinks />` to render both STX and BTC block height with links in all places where we want to show block details.
image
image
image

### DIFF
--- a/src/components/btc-stx-block-links.tsx
+++ b/src/components/btc-stx-block-links.tsx
@@ -1,0 +1,106 @@
+import { FC, Fragment } from 'react';
+import { FaBitcoin } from 'react-icons/fa';
+import { BsArrowRightShort } from 'react-icons/bs';
+import { Box, Text, color } from '@stacks/ui';
+import { css } from '@emotion/react';
+import { IconArrowRight } from '@tabler/icons';
+import { buildUrl } from '@components/links';
+import { useRouter } from 'next/router';
+import { StxInline } from '@components/icons/stx-inline';
+import { Circle } from '@components/circle';
+import { useAppSelector } from '@common/state/hooks';
+import { selectActiveNetwork } from '@common/state/network-slice';
+
+interface BtcStxBlockLinksProps {
+  btcBlockHeight?: number;
+  stxBlockHeight: number;
+  stxBlockHash: string;
+  fontSize?: string;
+}
+
+const wrapperStyle = css`
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+`;
+
+const iconStyle = css`
+  margin-right: 3px;
+`;
+
+const linkStyle = (secondary?: boolean, clickable?: boolean, fontSize?: string) => css`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  font-weight: 500;
+  font-family: 'Open Sauce', Inter, sans-serif;
+  ${fontSize ? `font-size: ${fontSize};` : 'font-size: 14px;'}
+  ${clickable &&
+  `
+    cursor: pointer;
+    &:hover {
+      text-decoration: underline;
+    }
+  `}
+  ${secondary &&
+  `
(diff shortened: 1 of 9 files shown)


src/components/btc-stx-block-links.tsx
```diff
@@ -0,0 +1,102 @@
...
+`;
+
+const linkStyle = (secondary?: boolean, clickable?: boolean, fontSize?: string) => css`
+  font-size: 14px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  font-weight: 500;
+  ${fontSize ? `font-size: ${fontSize};` : 'font-size: 14px;'}
+  ${clickable === true &&
```
beguene, line 38: This could just be 
```
clickable  &&
```

src/components/btc-stx-block-links.tsx
```diff
@@ -0,0 +1,102 @@
...
+      </Box>
+      {btcBlockHeight && (
+        <Fragment>
+          <IconArrowRight size="14px" color={'#747478'} css={arrowStyle} />
+          <FaBitcoin color={'#f7931a'} size={19} css={iconStyle} />
+          <Box
+            css={linkStyle(true, true)}
+            onClick={e => {
+              e.preventDefault();
+              window
```
beguene, line 91: There could be a security issue here as the third party website can have access to the `window.opener.location` and then change it to 'impersonate' our explorer.
The simpler way to fix this would be to use a <a> with a `rel="noopener"`, and it will also be a more semantic html.

```html
<a href="..." target="_blank" rel="noopener">

```

If it's really problematic to change it into a `<a>`, you can do this
```javascript
const win = window.open("XXX");
win.opener = null;
```

src/components/btc-stx-block-links.tsx
```diff
@@ -0,0 +1,102 @@
...
+  font-weight: 500;
+  ${fontSize ? `font-size: ${fontSize};` : 'font-size: 14px;'}
+  ${clickable === true &&
+  `
+    cursor: pointer;
+    &:hover {
+      text-decoration: underline;
+    }
+  `}
+  ${secondary === true &&
```
beguene, line 45: Same as above

beguene (changes requested):
Looks great @He1DAr !  🚀 
I just made a few comments.

src/components/btc-stx-block-links.tsx
He1DAr: unfortunately I can't use <a/> here because the parent is also <a/> (nested links are illegal)

src/components/btc-stx-block-links.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { selectActiveNetwork } from '@common/state/network-slice';
+
+interface BtcStxBlockLinksProps {
+  btcBlockHeight?: number;
+  stxBlockHeight: number;
+  stxBlockHash: string;
+  fontSize?: string;
+}
+
+const wrapperStyle = css`
```
andresgalante, line 21: I am a bit worried that we need to write so much custom CSS with hardcoded values. The component library should be able to hand something as this simple. 

Eventually, we'll need to reconsider our approach to the UI layer on the Explorer.

No action needed, just wanted to share my thought

He1DAr: let's discuss this in today's meeting

src/components/btc-stx-block-links.tsx
beguene: but you can still set the opener to null to prevent the third party from accessing the window.opener.location

He1DAr: I used `window.open(url,'_blank','noopener')`

beguene: My bad ! I did not see it. Awesome !

beguene (approved):

PR #812, opened 2022-08-09
feat: add fungible token images

beguene:
This adds images to the fungible tokens
I also 
- added unit test and gh action for ci
- refactored metadata fetch to use useQuery and leverage some caching
ft-images

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "lint:unused-exports": "ts-unused-exports tsconfig.json",
     "start": "next start",
     "typecheck": "tsc --noEmit",
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test",
+    "test:unit": "jest --forceExit ./src"
   },
   "version": "1.0.0",
   "author": "Hiro Systems PBC (formerly Blockstack PBC)",
@@ -106,6 +107,7 @@
     "unique-names-generator": "4.6.0",
     "use-constant": "1.1.0",
     "use-events": "1.4.2",
+    "valid-url": "1.0.9",
     "web-api-hooks": "3.0.2",
     "webpack": "5.47.0",
     "yup": "0.32.9"
@@ -122,6 +124,7 @@
     "@stacks/stacks-blockchain-api-types": "4.0.3",
     "@types/express": "4.17.13",
     "@types/fontfaceobserver": "0.0.6",
+    "@types/jest": "^28.1.6",
     "@types/just-debounce-it": "1.5.0",
     "@types/next": "9.0.0",
     "@types/node": "16.4.3",
@@ -131,6 +134,7 @@
     "@types/react": "17.0.15",
     "@types/store": "2.0.2",
     "@types/styled-system__theme-get": "5.0.1",
+    "@types/valid-url": "^1.0.3",
     "@typescript-eslint/eslint-plugin": "4.28.5",
     "@typescript-eslint/parser": "4.28.5",
     "@welldone-software/why-did-you-render": "6.2.0",
@@ -140,6 +144,7 @@
     "eslint-plugin-import": "2.23.4",
     "eslint-plugin-prettier": "3.4.0",
     "husky": "6.0.0",
+    "jest": "28.1.3",
     "lint-staged": "11.0.0",
     "package-json-cleanup-loader": "1.0.3",
     "prettier": "2.3.2",
@@ -165,6 +170,12 @@
         }
       ]
     ]
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "^@root(.*)$": "<rootDir>/src$1",
+      "^@common(.*)$": "<rootDir>/src/common$1"
+    }
   },
   "resolutions": {
     "eslint": "7.16.0",
--- a/src/common/hooks/use-token-metadata.ts
+++ b/src/common/hooks/use-token-metadata.ts
@@ -1,0 +1,25 @@
+import { getAssetNameParts } from '@common/utils';
+import { useApi } from '@common/api/client';
+import { useQuery } from 'react-query';
+
+export function useTokenMetadata(
+  token: string,
+  tokenType: 'non_fungible_tokens' | 'fungible_tokens'
+) {
+  const { address, contract, asset } = getAssetNameParts(token);
+  const { fungibleTokensApi } = useApi();
+  const T_24H_IN_MS = 1000 * 60 * 60 * 24;
+  return useQuery(
+    ['token-metadata', token],
+    () => {
+      const contractId = `${address}.${contract}`;
+      return fungibleTokensApi.getContractFtMetadata({
+        contractId,
+      });
+    },
+    {
+      staleTime: T_24H_IN_MS,
+      enabled: tokenType === 'fungible_tokens',
+    }
+  );
+}
--- a/src/common/utils/string-utils.spec.ts
+++ b/src/common/utils/string-utils.spec.ts
@@ -1,0 +1,15 @@
+import { convertUnicodeToAscii } from './string-utils';
+
+describe(convertUnicodeToAscii.name, () => {
+  it('should convert unicode to ascii', () => {
+    expect(convertUnicodeToAscii('Á')).toEqual('A');
+    expect(convertUnicodeToAscii('Û')).toEqual('U');
+    expect(convertUnicodeToAscii('ê')).toEqual('e');
+    expect(convertUnicodeToAscii('Ê')).toEqual('E');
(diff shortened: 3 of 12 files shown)


src/components/token-avatar.tsx
```diff
@@ -0,0 +1,29 @@
+import { DynamicColorCircle } from '@stacks/ui';
+import { FungibleTokenMetadata } from '@stacks/blockchain-api-client';
+
+import React from 'react';
+import { getAssetNameParts } from '@common/utils';
+import { imageCanonicalUriFromFtMetadata } from '@common/utils/token-utils';
+export function TokenAvatar({
```
He1DAr, line 7: can you define components with React.FC?
```
interface TokenAvatarProps {}
export const TokenAvatar: React.FC<TokenAvatarProps> = ({}) => ...
```

src/components/token-avatar.tsx
```diff
@@ -0,0 +1,29 @@
...
+}: {
+  token: string;
+  ftMetadata?: FungibleTokenMetadata | undefined;
+}) {
+  const { address, asset, contract } = getAssetNameParts(token);
+  const imageCanonicalUri = imageCanonicalUriFromFtMetadata(ftMetadata);
+  return imageCanonicalUri ? (
+    <TokenImage imageUri={imageCanonicalUri} />
+  ) : (
+    <DynamicColorCircle size="32px" mr="base" string={`${address}.${contract}::${asset}`}>
```
He1DAr, line 19: can you define this as a separate constant? maybe `<DefaultTokenImage/>`

src/components/token-avatar.tsx
```diff
@@ -0,0 +1,29 @@
...
+      {asset[0]}
+    </DynamicColorCircle>
+  );
+}
+
+const TokenImage = ({ imageUri }: { imageUri: string }) => {
+  return (
+    <img width={'36px'} height={'36px'} src={encodeURI(imageUri)} style={{ marginRight: '16px' }} />
```
He1DAr, line 27: maybe adding alt attribute

src/common/validation/validate-url.ts
```diff
@@ -0,0 +1,5 @@
+import { isWebUri } from 'valid-url';
+
+export function isValidUrl(str: string) {
```
He1DAr, line 3: thoughts on using a simple validation function instead of introducing a new library, something like:
```
const isValidUrl = (url: string) => {
  try {
    new URL(url);
  } catch (e) {
    return false;
  }
  return true;
};
```

src/common/utils/token-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+import { convertUnicodeToAscii } from './string-utils';
+
+export const isIconUrl = isValidUrl;
+
+export function isFtNameLikeStx(name: string): boolean {
+  return ['stx', 'stack', 'stacks'].includes(convertUnicodeToAscii(name).toLocaleLowerCase());
+}
+
+export function imageCanonicalUriFromFtMetadata(
+  meta: FungibleTokenMetadata | undefined
```
He1DAr, line 12:
```suggestion
  meta?: FungibleTokenMetadata
```

src/common/utils/token-utils.ts
```diff
@@ -0,0 +1,19 @@
+import { isValidUrl } from '@common/validation/validate-url';
+import { FungibleTokenMetadata } from '@stacks/blockchain-api-client';
+import { convertUnicodeToAscii } from './string-utils';
+
+export const isIconUrl = isValidUrl;
```
He1DAr, line 5: why do we need to assign the function to a new constant?

src/common/utils/token-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function isFtNameLikeStx(name: string): boolean {
+  return ['stx', 'stack', 'stacks'].includes(convertUnicodeToAscii(name).toLocaleLowerCase());
+}
+
+export function imageCanonicalUriFromFtMetadata(
```
He1DAr, line 11: could be a unit test candidate

src/common/utils/token-utils.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(isFtNameLikeStx('Stácks')).toBeTruthy();
+    expect(isFtNameLikeStx('Stáçks')).toBeTruthy();
+    expect(isFtNameLikeStx('stocks')).toBeFalsy();
+    expect(isFtNameLikeStx('miamicoin')).toBeFalsy();
+    expect(isFtNameLikeStx('')).toBeFalsy();
+  });
+});
+
+describe(isIconUrl.name, () => {
+  it('detect valid icon url', () => {
```
He1DAr, line 18: this will pass for non-image URLs too?

src/common/utils/string-utils.ts
```diff
@@ -0,0 +1,3 @@
+export function convertUnicodeToAscii(str: string): string {
```
He1DAr, line 1: this is to remove accents from accentuated characters? maybe rename it to reflect that
```
removeAccentsFromString
```

src/common/utils/string-utils.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { convertUnicodeToAscii } from './string-utils';
+
+describe(convertUnicodeToAscii.name, () => {
+  it('should convert unicode to ascii', () => {
```
He1DAr, line 4:
```suggestion
  it('should remove accents from string', () => {
```

src/common/hooks/use-token-metadata.ts
```diff
@@ -0,0 +1,25 @@
...
+import { useApi } from '@common/api/client';
+import { useQuery } from 'react-query';
+
+export function useTokenMetadata(
+  token: string,
+  tokenType: 'non_fungible_tokens' | 'fungible_tokens'
+) {
+  const { address, contract, asset } = getAssetNameParts(token);
+  const { fungibleTokensApi } = useApi();
+  const T_24H_IN_MS = 1000 * 60 * 60 * 24;
```
He1DAr, line 11: we should probably define these settings in a common place and use them in other queries, we can tackle this in a separate PR.

package.json
```diff
@@ -140,6 +144,7 @@
     "eslint-plugin-import": "2.23.4",
     "eslint-plugin-prettier": "3.4.0",
     "husky": "6.0.0",
+    "jest": "^28.1.3",
```
He1DAr, line 147: we usually use exact version `"28.1.3"`

He1DAr (changes requested):
Great work Beguene! left a few comments.

src/common/validation/validate-url.ts
kyranjamie: `new URL("javascript:alert('lskdjfs')")` doesn't throw ⚠️

src/common/utils/token-utils.ts
beguene: It was just for semantic, it's not relevant anymore as I changed the logic

src/common/utils/token-utils.spec.ts
beguene: Added a test and a check for non-images

src/common/utils/string-utils.ts
beguene: I think the name is correct here. This is not only for accents, it's just converting unicode to ascii so it's always readable and prevent spoofing.

src/common/validation/validate-url.ts
beguene: Indeed, we should do more than just using new URL() as @kyranjamie pointed out.
I introduce third party libraries with care and I think it's justified here as it has no dependencies, is small and easily [verifiable](https://github.com/ogt/valid-url/blob/master/index.js), has a single purpose and the url format in general does not change.

src/common/utils/token-utils.ts
beguene: Added some  tests

src/components/token-avatar.tsx
beguene: I think React.FC here is not useful, this component doesn't even use any children. I prefer in general to be explicit in the props.

He1DAr: what are your thoughts on always defining props as a separate interface on top of the component?

He1DAr: you did it already, nice!

He1DAr (approved):

src/components/token-avatar.tsx
beguene: Yes I like it too, thanks for pointing it out !

PR #813, opened 2022-08-09
feat: add filters to homepage tx list

### LINKED ISSUE #355: Add filters to the transaction list on the homepage
Is there a feature already in the works for this?

He1DAr:
Add filters to the homepage tx list
image
closes #355

### DIFF
--- a/src/features/transaction-list.tsx
+++ b/src/features/transaction-list.tsx
@@ -7,11 +7,12 @@ import type {
 } from '@store/transactions';
 import type { MempoolTransaction, Transaction } from '@stacks/stacks-blockchain-api-types';
 import { useFilterState } from '@common/hooks/use-filter-state';
-import { TxFilterTypes } from '@features/transactions-filter/transactions-filter-slice';
+import { useFilterScope } from '@features/transactions-filter/hooks/use-filter-scope';
 
 interface TransactionListProps {
   isLastPage?: boolean;
   data?: TransactionsListResponse | MempoolTransactionsListResponse;
+  mempoolSelected?: boolean;
 }
 
 function getUniqueListBy<T>(arr: T[], key: keyof T): T[] {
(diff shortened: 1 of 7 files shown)


src/features/transaction-list.tsx
```diff
@@ -27,8 +28,13 @@ export const TransactionList = (props: TransactionListProps) => {
         : undefined,
     [data]
   );
+  const router = useRouter();
+  const filterScope =
```
beguene, line 32: We could maybe extract this logic as it's the same as the one above and we might extend this later to other pages.

beguene:
@He1DAr All the filtering works fine but there is an issue when we filter on the 'confirmed' tab and then goes to the 'pending' tab'. The 'Filter' label disappears.

https://user-images.githubusercontent.com/1501454/183823446-b225131f-4db3-48c2-a8b3-0a46ab042ea4.mov

We could at least show the 'Filter' component in the pending tab.

Even better would be to change the colour of the 'Filter' label so the user knows if the filtering is active. Or any other kind of visual cues to convey this. This is not related to this task but a nice to have.

He1DAr:
> @He1DAr All the filtering works fine but there is an issue when we filter on the 'confirmed' tab and then goes to the 'pending' tab'. The 'Filter' label disappears.
> 
>  home-page-filter.mov 
> We could at least show the 'Filter' component in the pending tab.
> 
> Even better would be to change the colour of the 'Filter' label so the user knows if the filtering is active. Or any other kind of visual cues to convey this. This is not related to this task but a nice to have.

It looks like filters are skipped for pending tx list in all places, I fixed the issue with pending tx not showing but I think we should keep it consistent with the other tx pages and maybe update it in all pages if we wanna show it for pending tx?

He1DAr:
We definitely need to surface selected filters as you suggested, I'll create an issue for it.

beguene (approved):

PR #825, opened 2022-08-18
feat: add stx and btc prices to header

He1DAr:
Add stx and btc prices to header

image


beguene (approved):
Looks great @He1DAr ! 🔥

PR #828, opened 2022-08-22
chore: reduce container size

deantchi:
Use multistage dockerfile build.  Source code is now excluded from the final image.  
There will be some more build options to further reduce the size once the project has been updated to use NextJS v7.x
As of now, the total container size has been reduced about 50%


CharlieC3 (commented):
Looks good, my only concern is the ENV vars created in the build stage will not be available at run time in the second stage.

@He1DAr correct me if I'm wrong, but I believe at least some of these env vars are needed at run time, though I'm not sure if all are.

I would suggest looking into setting the ENV vars in a base stage, then using that base stage for the final stage [like here](https://github.com/moby/moby/issues/37345#issuecomment-400250849).

deantchi:
hi @He1DAr any thought on what Charlie mentioned. Thanks

He1DAr:
> hi @He1DAr any thought on what Charlie mentioned. Thanks

Yep, some vars are needed for runtime https://github.com/hirosystems/explorer/blob/main/next.config.js

deantchi:
@CharlieC3 it doesn't look like you can persist `ENV VAR` directives through a multi stage build.  However `ARGs` can persist so I moved it to the global level.

CharlieC3:
@deantchi I think some or all of the env vars are needed at both build-time and run-time.

CharlieC3 (approved):
LGTM! :shipit:

PR #832, opened 2022-08-23
feat: display tx transfers for addresses

He1DAr:
Display tx transfers for addresses

image


beguene (approved):
Awesome @MrHeidar ! 🚀

PR #838, opened 2022-08-25
fix: refactor microblock to fetch client side

### LINKED ISSUE #626: Disable server-side API requests
The Explorer's backend is performing a high volume of requests to the API and is a blocker for adjusting our API rate limiting. 

AFAIK this is a nextjs server-side rendering process which needs to be tweaked.

### LINKED ISSUE #834: Move <RecoilRoot> out of the main _app.tsx to only the sandbox page
Currently ```<RecoilRoot>``` is placed in out _app.tsx wrapping all inner components for every pages.
But recoil is used only in sandbox page. We could reduce the bundle size for all the pages except the sandbox by remove this ```<RecoilRoot>``` and placing it only for the sandbox page.

As per my first testing this will free up ~20k of js on first load for all pages

beguene:
closes #626 closes #834

Split fetchTransaction used in many pages into 2 distinct fetches : one of the tx and one for the block so we can leverage caching, parallelism and improve perf
added eslint rule for react hooks
remove some unused packages
Moved RecoilRoot> to sandbox only to reduce the size of first load js on all pages (except sandbox)
Remove some dead code
Remove some jotai-query toolkit dependencies (some are still left in sandbox)
remove AtomDebug of jotai from the _app

By removing Recoil we shaved off 24kb on first load js

Before
b

After
a


src/components/loaders/skeleton-transaction.tsx
```diff
@@ -216,15 +216,18 @@ export const SkeletonTransactionSummary = () => {
...
-        <SkeletonTransaction key={i} />
-      </div>
-    ))}
-  </>
-);
+export const SkeletonTransactionList = ({ length }: { length?: number }) => {
+  const num = length ?? 10;
+  return (
+    <>
+      {new Array(num).fill(true).map((_, i) => (
```
He1DAr, line 223: can do `[...Array(num)].map`

He1DAr (approved):
Nice work! 🔥

PR #839, opened 2022-08-26
feat: add principal validation to sandbox

He1DAr:
image

### DIFF
--- a/src/sandbox/components/screens/call-functions/components/available-functions.tsx
+++ b/src/sandbox/components/screens/call-functions/components/available-functions.tsx
@@ -5,7 +5,7 @@ import { useRecoilValue, useRecoilState, useSetRecoilState } from 'recoil';
 import { Input } from '@components/inputs';
 import { useFormik } from 'formik';
 import { Button } from '@components/button';
-import { border } from '@common/utils';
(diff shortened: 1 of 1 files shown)


src/sandbox/components/screens/call-functions/components/available-functions.tsx
```diff
@@ -177,8 +183,26 @@ const FunctionSingleView = () => {
...
+    validateOnBlur: false,
+    validate: values => {
+      const errors = {};
+      Object.keys(values).forEach(key => {
+        const type = fn.args.find(({ name }) => name === key).type;
+        const optionalType = type?.optional;
+        if (type === 'principal' || (optionalType === 'principal' && !!values[key])) {
+          const validPrincipal = validateStacksAddress(values[key]);
+          if (!validPrincipal) {
+            errors[key] = 'Invalid Stacks address.';
```
sabbyanandan, line 200: A nitpick; perhaps remove the `.` from the error message (haha). If that is the format followed today for all messages, please keep it as-is.

He1DAr: Yep we have the same exact message in other places🙂, we should revisit this at some point.

andresgalante (approved):

PR #843, opened 2022-08-30
feat: upgrade next.js and use standalone output

He1DAr:
- upgrade next.js to v12.2.5
- use [standalone output](https://nextjs.org/docs/advanced-features/output-file-tracing)


beguene (approved):
Nice one @MrHeidar and thanks for the build fix 😄

PR #844, opened 2022-08-30
reduce container size further

deantchi:
thanks to @He1DAr 's upgrade, we can now reduce this container size down to ~176mb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,8 +53,8 @@ WORKDIR /app
 COPY --from=build /app/next.config.js /app/next.config.js
 COPY --from=build /app/public /app/public
 COPY --from=build /app/package.json /app/package.json
-COPY --from=build --chown=nextjs:nodejs /app/.next/standalone ./
-COPY --from=build --chown=nextjs:nodejs /app/.next/static ./.next/static
+COPY --from=build /app/.next/standalone /app
+COPY --from=build /app/.next/static /app/.next/static
 
 EXPOSE 3000
 CMD [ "node", "server.js" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,9 +52,9 @@ WORKDIR /app
 
 COPY --from=build /app/next.config.js /app/next.config.js
 COPY --from=build /app/public /app/public
-COPY --from=build /app/package.json /app/package.json
-COPY --from=build /app/.next/standalone /app
 COPY --from=build /app/.next/static /app/.next/static
+COPY --from=build /app/.next/standalone /app
+
 
 EXPOSE 3000
 CMD [ "node", "server.js" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,12 +50,11 @@ RUN apk --no-cache add --virtual \
 
 WORKDIR /app
 
-COPY --from=build /app/node_modules /app/node_modules
-COPY --from=build /app/.next /app/.next
 COPY --from=build /app/next.config.js /app/next.config.js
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -50,12 +50,11 @@ RUN apk --no-cache add --virtual \
 
 WORKDIR /app
 
-COPY --from=build /app/node_modules /app/node_modules
-COPY --from=build /app/.next /app/.next
 COPY --from=build /app/next.config.js /app/next.config.js
 COPY --from=build /app/public /app/public
 COPY --from=build /app/package.json /app/package.json
-COPY --from=build /app/yarn.lock /app/yarn.lock
+COPY --from=build --chown=nextjs:nodejs /app/.next/standalone ./
```
CharlieC3, line 56: Why is `chown` used here but not for the other COPY directives? Is it needed here?

Dockerfile
```diff
@@ -50,12 +50,11 @@ RUN apk --no-cache add --virtual \
...
+COPY --from=build --chown=nextjs:nodejs /app/.next/static ./.next/static
 
 EXPOSE 3000
-CMD [ "yarn", "start" ]
+CMD [ "node", "server.js" ]
```
CharlieC3, line 60: Are we moving away from yarn? Why not start the explorer using the [pre-defined package.json script](https://github.com/hirosystems/explorer/blob/main/package.json#L14)?

CharlieC3 (commented):
Left some questions!

Dockerfile
CharlieC3: Also, can we specify the full destination path like the other directives? (`./` vs `/app/.....`)

Dockerfile
deantchi: the `standalone` feature produces a minimal file `server.js` which is used to run and it doesn't look like `yarn` can execute this.  Additionally, per their documentation, the `public` and `static` assets should be served by a CDN instead of baking it into the image.  This is fine for now but something definitely worth looking into in the future.

Dockerfile
deantchi: I've tested again, `--chown` is not required, so I will be removing this.  Also, changed the path to the full path.

Dockerfile
CharlieC3: Interesting, thanks for clarifying!

Dockerfile
```diff
@@ -50,12 +50,11 @@ RUN apk --no-cache add --virtual \
 
 WORKDIR /app
 
-COPY --from=build /app/node_modules /app/node_modules
-COPY --from=build /app/.next /app/.next
 COPY --from=build /app/next.config.js /app/next.config.js
 COPY --from=build /app/public /app/public
 COPY --from=build /app/package.json /app/package.json
-COPY --from=build /app/yarn.lock /app/yarn.lock
+COPY --from=build /app/.next/standalone /app
```
CharlieC3, line 56: Not sure if this was intentional, but I believe this is going to wipe the destination `/app` dir and replace it with the contents of the source `/app/.next/standalone` dir.
Did you mean `COPY --from=build /app/.next/standalone /app/.next/standalone` instead?

deantchi: the `standalone` directory is the output of this new feature.  one can run the application directly from the `/app/.next/standalone` directory, or copy it out into another location and run it from there.

to your point, it does not wipe the destination, it only affects duplicate files and it looks like the line #57 redundant so I will remove that.

CharlieC3 (approved):
:shipit:

Issue #826, opened 2022-08-22
Add a link to the blockchain.com Explorer from the Bitcoin Block Hight on the block page


andresgalante:
The block page has information about the bitcoin anchor and its block height.
<img width="346" alt="Screen Shot 2022-08-22 at 10 59 16" src="https://user-images.githubusercontent.com/1832037/185939509-d520535c-0f89-4350-a54d-3462e63caf5f.png">

This information is currently static and it'd be better if we add a link for users to explore more information about the block.

The task is to add a link to blockchain.com from the BTC block height text.

PR #845, opened 2022-09-01
feat: add a link to the blockchain.com explorer from the bitcoin block height on the block page

leahjlou:
Closes https://github.com/hirosystems/explorer/issues/826

Within the "Bitcoin anchor" section on the right side of the block page, the "Bitcoin block height" text will have a link to the blockchain.com explorer.

Screen Shot 2022-08-31 at 4 26 13 PM

### DIFF
--- a/src/components/btc-anchor-card.tsx
+++ b/src/components/btc-anchor-card.tsx
@@ -23,7 +23,15 @@ export const BtcAnchorBlockCard: React.FC<FlexProps & { block: Block }> = ({ blo
               label: {
                 children: 'Bitcoin block height',
               },
-              children: `#${block.burn_block_height}`,
+              children: (
+                <Link
+                  as="a"
+                  target="_blank"
+                  href={`https://www.blockchain.com/btc/block/${block.burn_block_height}`}
+                >
+                  #{block.burn_block_height}
+                </Link>
+              ),
             },
             {
               label: {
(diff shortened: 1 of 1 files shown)


src/components/btc-anchor-card.tsx
```diff
@@ -23,7 +23,15 @@ export const BtcAnchorBlockCard: React.FC<FlexProps & { block: Block }> = ({ blo
               label: {
                 children: 'Bitcoin block height',
               },
-              children: `#${block.burn_block_height}`,
+              children: (
+                <Link
+                  as="a"
+                  target="_blank"
+                  href={`https://www.blockchain.com/btc/block/${block.burn_block_height}?utm_source=stacks_explorer`}
```
leahjlou, line 30: Included the `utm_source` param here as in the other links, but I am not sure exactly how it's used by the blockchain.com explorer. I noticed that the param is not included in some other links (added here: https://github.com/hirosystems/explorer/pull/811). Any thoughts on whether `utm_source` should be added to the `<BtcStxBlockLinks />` component?

beguene: I think we should not add `utm_source` here which is commonly used for marketing campaigns. We are not responsible for blockchain.com marketing and it might even mess up their analytics. They can always know the source by using the `window.document.referrer`

beguene (approved):
Nice first PR @leahjlou ! 🚀  (just made a comment about removing the `utm_source`)

He1DAr (approved):
LGTM!

PR #846, opened 2022-09-01
chore: add husky hook to lint commit messages

leahjlou:
The [build process for this project runs `commitlint`](https://github.com/hirosystems/explorer/blob/main/.github/workflows/commitlint.yml) to standardize commit messages, but it'd be nice for developers to know if they've typed an invalid commit message right away, before it is pushed and building. This change adds a husky hook to lint commit messages as they're being authored, and prevents invalid commits from being created.

For example, here is what a developer may see if they try to commit with an invalid message:
Screen Shot 2022-09-01 at 2 14 33 PM

If you type a valid commit message, the commit is added normally:
Screen Shot 2022-09-01 at 2 14 52 PM

Note: after this change is merged, any cloned instance of the project will need to run `yarn install` for this hook to work. On initial setup of the project, it should work automatically.


beguene (approved):
Great idea !

PR #855, opened 2022-09-07
feat: show usd values alongside balances on address page

### LINKED ISSUE #853: Show current STX USD value next to account balance
On the address page

leahjlou:
Closes #853 

Shows USD price alongside all balances on the address page.

Screen Shot 2022-09-07 at 5 40 15 PM

Screen Shot 2022-09-07 at 5 39 30 PM

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn lint:prettier
--- a/src/components/balances/stx-balance-card.tsx
+++ b/src/components/balances/stx-balance-card.tsx
@@ -18,21 +18,41 @@ import { useAtomValue } from 'jotai/utils';
 import { accountInViewTokenOfferingData } from '@store/currently-in-view';
 import { MODALS } from '@common/constants';
 import { useAppDispatch } from '@common/state/hooks';
+import { useCurrentStxPrice } from '@common/hooks/use-current-prices';
+
+const usdFormatter = new Intl.NumberFormat('en-US', {
+  style: 'currency',
+  currency: 'USD',
+});
 
 export const BalanceItem = ({ balance, ...rest }: any) => {
+  const { data: stxPrice } = useCurrentStxPrice();
+
+  const formattedBalance = balance.toLocaleString(undefined, {
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 6,
+  });
+  const usdBalance = usdFormatter.format(balance * stxPrice);
   const localeDecimalSeparator = getLocaleDecimalSeparator();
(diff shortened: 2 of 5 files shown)


src/components/balances/stx-balance-card.tsx
```diff
@@ -18,21 +18,36 @@ import { useAtomValue } from 'jotai/utils';
...
+
+const usdFormatter = new Intl.NumberFormat('en-US', {
+  style: 'currency',
+  currency: 'USD',
+});
 
 export const BalanceItem = ({ balance, ...rest }: any) => {
+  const { data: stxPrice } = useCurrentStxPrice();
+
+  const formattedBalance = balance.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 6 });
```
leahjlou, line 31: Changed this component to use the raw number for the `balance` prop rather than the formatted string for the most accuracy when calculating USD value

src/common/utils/market-utils.ts
```diff
@@ -0,0 +1,19 @@
+import { useQuery } from 'react-query';
```
beguene, line 1: Great that you extracted all the logic. The current folder organisation is not great, but you could maybe place this file under `common/hooks/` instead.

beguene (commented):
Code looks good to me ! 
(You still have lint check failing https://github.com/hirosystems/explorer/runs/8239362059?check_suite_focus=true)

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn lint:prettier
```
leahjlou, line 4: Adding precommit formatting check

andresgalante:
Can we align the stacks icon top instead of center? Other than that 🚀

He1DAr (approved):

leahjlou:
> Can we align the stacks icon top instead of center? Other than that 🚀
> 
> `align-items: flex-start;` instead of `center`
> 
> Screen Shot 2022-09-08 at 13 48 48

I like that 👍

.husky/pre-commit
beguene: @leahjlou I think running this at every commit for all developers can become annoying. Imagine a developer commits several times intermediate work, we don't want to run this as it's not running fast and will slow down as the number of files increases. We should configure it at the editor level or the developer can still add this as a personal git hook

leahjlou: Sorry @beguene I forgot to address this comment before I merged this PR. I agree with your point, opened a new PR here to remove it https://github.com/hirosystems/explorer/pull/859

beguene: No issue @leahjlou I actually made my comment after the merge. Thanks for fixing it.

PR #862, opened 2022-09-15
feat: fetch txs and blocks on focus

He1DAr:
Fetch txs and blocks on focus

### DIFF
--- a/src/features/account-transaction-list.tsx
+++ b/src/features/account-transaction-list.tsx
@@ -28,13 +28,13 @@ export const AccountTransactionList: React.FC<{ contractId: string }> = ({ contr
     transactionQK(TransactionQueryKeys.transactionsWithTransfersForAddressInfinite, contractId),
     ({ pageParam }) =>
       queries.fetchTransactionsWithTransfersForAddress(contractId, undefined, pageParam || 0)(),
-    { getNextPageParam, enabled: !!contractId }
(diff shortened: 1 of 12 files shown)


src/features/account-transaction-list.tsx
```diff
@@ -28,13 +28,13 @@ export const AccountTransactionList: React.FC<{ contractId: string }> = ({ contr
     transactionQK(TransactionQueryKeys.transactionsWithTransfersForAddressInfinite, contractId),
     ({ pageParam }) =>
       queries.fetchTransactionsWithTransfersForAddress(contractId, undefined, pageParam || 0)(),
-    { getNextPageParam, enabled: !!contractId }
+    { getNextPageParam, enabled: !!contractId, refetchOnWindowFocus: true }
```
leahjlou, line 31: Are these all needed? I thought `refetchOnWindowFocus` was `true` by default

aulneau: https://github.com/hirosystems/explorer/blob/main/src/pages/_app.tsx#L43

it's set to default off here btw

leahjlou (approved):

src/features/account-transaction-list.tsx
He1DAr: I wanted to only enable it for txs/blocks and keep it disabled for the rest or requests.

PR #884, opened 2022-10-07
feat: show events with pagination (load more)

### LINKED ISSUE #866: Show all events using pagination
This tx https://explorer.stacks.co/txid/0xffe51f5fddceb61b2b2f36e72ce8370a0dbcdfdfc96dd784327fcaf9d2367eac?chain=mainnet
has 207 events but we shows only a constant subset of it
This task is to add a load more or pagination so the user can see all the events
The endpoint https://docs.hiro.so/api#tag/Transactions/operation/get_transaction_by_id used here already supports pagination.

beguene:
closes #866
We could use this direct endpoint https://docs.hiro.so/api#tag/Transactions/operation/get_filtered_events  but there is a small issue with the api sdk and the response values are not meaningful (see issue for the details)
https://github.com/hirosystems/explorer/issues/866

I ended up using the /tx endpoint and paginate with `event_limit` and `event_offset` and formatting the response to look like a paginated response `ApiResponseWithResultsOffset`

[events](https://user-images.githubusercontent.com/1501454/194539875-59923a6f-43c3-4e54-862f-fc7153c66546.mov)

For testing and this PR I made the limit = 50 instead of 200 so we can easily test. I will revert to 200 before merge.

 Good tx to test
[https://explorer.stacks.co/txid/0xffe51f5fddceb61b2b2f36e72ce8370a0dbcdfdfc96dd784327fcaf9d2367eac?](https://explorer.stacks.co/txid/0xffe51f5fddceb61b2b2f36e72ce8370a0dbcdfdfc96dd784327fcaf9d2367eac?chain=mainnet)


src/common/api/utils.ts
```diff
@@ -24,3 +24,11 @@ export const getNetworkModeFromNetworkId = (networkId: ChainID) => {
       return undefined;
   }
 };
+
+export const constructQueryParams = (obj: Record<string, string | number | boolean>): string => {
```
leahjlou, line 28: I don't see this function called here, is it needed?

src/components/tx-events.tsx
```diff
@@ -306,27 +317,86 @@ const Item: React.FC<{ event: TransactionEvent; isLast?: boolean }> = ({ event,
...
-  txId: string;
   events: TransactionEvent[];
+  txId: string;
+  event_count: number;
 } & FlexProps) => {
-  return events?.length ? (
+  const queries = useTransactionQueries();
+  // fetch extra events if there are more than the default returned by the fetch tx api
+  // and display them below
+  const { data, ...actions } = useInfiniteQuery(
```
leahjlou, line 361: This might be worth extracting to a separate file i.e. `useTxEvents.ts`, the component is pretty large

src/features/transaction/use-transaction-queries.ts
```diff
@@ -47,28 +48,59 @@ export const getTransactionQueries = (networkUrl: string) => {
...
+      () =>
+        fetchTransactionWithContract({ txId, eventLimit, eventOffset }),
+    fetchEventsForTx:
+      ({
+        txId,
+        eventLimit = DEFAULT_TX_EVENTS_LIMIT,
+        eventOffset = 0,
+      }: GetTransactionByIdRequest) =>
+      async () => {
+        console.log('foo events');
```
leahjlou, line 93: Stray console.log, should be removed

src/common/api/utils.ts
beguene: I was using it before but this is not used anymore. I will remove it.

leahjlou (approved):

PR #893, opened 2022-10-20
feat: display bns names

### LINKED ISSUE #890: Display BNS name for address
Instead of just "names: 1" under collectibles, so users can see the actual name registered for the address:

image

He1DAr:
closes #890 and #897

### DIFF
--- a/src/components/balances/principal-token-balances.tsx
+++ b/src/components/balances/principal-token-balances.tsx
@@ -1,12 +1,21 @@
 import * as React from 'react';
-import { AddressBalanceResponse } from '@stacks/stacks-blockchain-api-types';
+import {
+  AddressBalanceResponse,
+  NonFungibleTokenHoldingsList,
+} from '@stacks/stacks-blockchain-api-types';
 import { Box, color, FlexProps, Grid, GridProps } from '@stacks/ui';
 import { Caption } from '@components/typography';
 import { Section } from '@components/section';
 import { TokenAssetListItem } from '@components/balances/token-asset-list-item';
 import { HoverableItem } from '@components/hoverable';
+import { cvToJSON, hexToCV } from '@stacks/transactions';
+import { useMemo } from 'react';
+import { hexToString } from '@common/utils';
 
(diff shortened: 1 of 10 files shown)


src/components/balances/principal-token-balances.tsx
```diff
@@ -57,14 +71,33 @@ const tabs: { label: string; slug: 'ft' | 'nft' }[] = [
...
   const TabContent = activeTab === 'ft' ? FtBalances : NftBalances;
 
+  const bnsHexValues = useMemo(
+    () =>
+      nftHoldings?.results
+        ?.filter(nftHolding => nftHolding.asset_identifier.endsWith('.bns::names'))
+        ?.reduce((acc, data) => {
+          acc[data.asset_identifier] = data.value?.hex
+            ? {
+                name: hexToString(
```
leahjlou, line 88: `hexToString` doesn't handle `undefined` inputs well, and the optional chaining here suggests it's possible for the input to be undefined -- maybe there should be a nice default display value to be on the safe side

src/features/search/use-search.ts
```diff
@@ -12,16 +13,37 @@ import { selectSearchTerm } from '@features/search/search-slice';
 import { useDebounce } from '@common/hooks/use-debounce';
 import { isNumeric } from '@common/utils';
 import { Block } from '@stacks/blockchain-api-client';
+import { BTC_BNS_CONTRACT } from '@common/constants';
+import { cvToHex, tupleCV, bufferCVFromString } from '@stacks/transactions';
 
 export const useSearchQuery = (id: string) => {
-  const { searchApi, blocksApi } = useAtomValue(apiClientsState);
+  const { searchApi, blocksApi, nonFungibleTokensApi } = useAtomValue(apiClientsState);
+  const isBtcName = id.endsWith('.btc');
```
leahjlou, line 21: Not a necessity, but it might be nice to make this case insensitive (and down on lines 34-35)

leahjlou (approved):

src/features/search/use-search.ts
He1DAr: we decided to leave it case-sensitive to match the actual name

PR #901, opened 2022-10-25
feat: display hardcoded miner upgrade notification

saralab:
Hardcoded message for miners to upgrade to 2.05.0.5.1

### DIFF
--- a/src/features/status-bar/status-bar.tsx
+++ b/src/features/status-bar/status-bar.tsx
@@ -39,16 +39,16 @@ const iconStyle = css`
 
 export const StatusBar: React.FC = () => {
   const [status, setStatus] = useState<StatusProps>({
-    description: '',
-    indicator: Indicator.none,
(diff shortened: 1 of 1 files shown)


src/features/status-bar/status-bar.tsx
```diff
@@ -39,16 +39,11 @@ const iconStyle = css`
 
 export const StatusBar: React.FC = () => {
   const [status, setStatus] = useState<StatusProps>({
-    description: '',
-    indicator: Indicator.none,
+    description: 'Stacks Miners — Upgrade to  2.05.0.5.1 NOW!',
+    indicator: Indicator.critical,
   });
-  useEffect(() => {
-    void fetch('https://status.hiro.so/api/v2/status.json')
```
diwakergupta, line 46: Don't we want to still have this around? Seems like a loss in functionality?

Issue #902, opened 2022-10-25
Show error for non-existent transaction


obycode:
**Describe the bug**
I noticed this morning that if I attempt to directly open a txid that does not exist, instead of showing me some error saying that, it just looks like it is loading forever. For example: https://explorer.stacks.co/txid/0xbb7bcc79a698c665edcd01b934967b012c5329ba6b88356f4bf50ab0d9551e61?chain=mainnet

**To Reproduce**
1. Go to https://explorer.stacks.co/txid/0xbb7bcc79a698c665edcd01b934967b012c5329ba6b88356f4bf50ab0d9551e61?chain=mainnet
2. See the placeholder images and no error messages

**Expected behavior**
I would expect to see some error indicating that the transaction is not found -- something similar to what shows if I search for an unknown txid ("No block or transaction found with hash ...")

**Desktop (please complete the following information):**
 - OS: macOS 12.6
 - Browser: Safari, Chrome
 - Version: Safari 16.0, Chrome 106.0.5249.119

PR #910, opened 2022-11-03
fix: add error check for non-existent transaction on txid page

tuanphungcz:
fixes https://github.com/hirosystems/explorer/issues/902

## Issue
`useQuery` fetches the non-existent tx several times, but the component doesn't handle the error state so it looks like it's loading forever.

## Solution
added an error check so that after a few failed fetching attempts it would display the `TxNotFound` component.

## Demo
https://user-images.githubusercontent.com/16448321/199724219-4b792288-1ade-4759-a1e7-588721595840.mp4

local test url: http://localhost:3000/txid/0xbb7bcc79a698c665edcd01b934967b012c5329ba6b88356f4bf50ab0d9551e61?chain=mainnet

### DIFF
--- a/src/pages/txid/[txid].tsx
+++ b/src/pages/txid/[txid].tsx
@@ -29,13 +29,13 @@ const TransactionPage: NextPage<TransactionPageProps> = ({ error, isPossiblyVali
     staleTime: 2000,
   };
 
-  const { data: tx } = useQuery(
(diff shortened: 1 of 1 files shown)


andresgalante:
@He1DAr would you be able to review this PR, please?

He1DAr (approved):

PR #912, opened 2022-11-04
fix: refactor sandbox

He1DAr:
- refactor sandbox sections into Next.js pages
- remove jotai, jotai-query-toolkit and recoil
- upgrade @stacks/* libraries
- remove ts-ignore and fix types


package.json
```diff
@@ -38,8 +40,9 @@
     "@stacks/auth": "4.0.1",
     "@stacks/blockchain-api-client": "4.0.1",
     "@stacks/common": "4.1.0",
-    "@stacks/connect": "6.8.4",
-    "@stacks/connect-ui": "5.5.0",
+    "@stacks/connect": "7.0.0",
+    "@stacks/connect-react": "^20.0.0",
```
leahjlou, line 44: Should we pin this?

src/features/search/items/address-result-item.tsx
```diff
@@ -1,23 +1,30 @@
...
 export const AddressResultItem: React.FC<AddressResultItemProps> = ({ result }) => {
-  if (!result || !result.found || result.result.entity_type !== 'standard_address') return null;
   const principal = result.result.entity_id;
-  const displayName = result.result.display_name;
+  const displayName =
+    result.result.entity_type === 'standard_address' ? result.result.display_name : '';
   const truncatedPrincipal = truncateMiddle(principal, 4);
-  const stx = useAtomValue(accountStxBalanceResponseState(principal));
+  const { accountsApi } = useApi();
+  const { data: stxBalance } = useQuery(
```
leahjlou, line 23: just an idea: pulling this query out into its own file would make it more reusable and help whoever might be needing it in the future to make sure they use the same query key

src/features/transactions-filter/hooks/use-filter-scope.ts
```diff
@@ -7,7 +7,10 @@ const pathnameFilterScopeMap: Record<string, TxFilterTypes> = {
   '/address/[principal]': TxFilterTypes.AddressTxFilter,
   '/block/[hash]': TxFilterTypes.BlocksPageTxFilter,
   '/microblock/[hash]': TxFilterTypes.BlocksPageTxFilter,
-  '/sandbox/[...param]': TxFilterTypes.SandboxTxFilter,
+  '/sandbox/deploy': TxFilterTypes.SandboxTxFilter,
```
leahjlou, line 10: 🎉

src/modules/sandbox/components/SideNav.tsx
```diff
@@ -1,66 +1,72 @@
...
     url: string;
     icon: any;
     isDisabled?: boolean;
+    isSelected: boolean;
   }[] = [
     {
       label: 'Write & Deploy Contracts',
       url: buildUrl(`/sandbox/deploy`, networkMode),
       icon: <ClarityIcon size="24px" />,
+      isSelected: router.pathname === '/sandbox/deploy',
```
leahjlou, line 30: I'm quite sure this is the case, but did you confirm that `router.pathname` does _not_ include the query string if it's present?

src/modules/sandbox/hooks/useUser.ts
```diff
@@ -0,0 +1,69 @@
...
+  const [mounted, setMounted] = useState(false);
+  const queries = useAddressQueries();
+
+  useEffect(() => setMounted(true), []);
+
+  useEffect(() => {
+    if (userSession.isUserSignedIn()) {
+      dispatch(setUserData({ userData: userSession.loadUserData() }));
+    }
+  }, []);
```
leahjlou, line 30: Does this work when the user signed in status changes? I'm wondering if the return value of `userSession.isUserSignedIn()` should maybe be watched in the dependency array of this `useEffect`

src/modules/sandbox/sandbox-slice.ts
```diff
@@ -0,0 +1,96 @@
...
+      }>
+    ) => {
+      const defaultOptions: AuthOptions = {
+        appDetails: APP_DETAILS,
+        userSession: state.userSession,
+      };
+      showConnect(Object.assign(defaultOptions, action.payload.authOptions || {}));
+    },
+    disconnect: state => {
+      console.log('[debug] disconnect');
```
leahjlou, line 42: Is this console.log intended to be kept in here?

src/modules/sandbox/components/SideNav.tsx
```diff
@@ -1,66 +1,72 @@
...
     },
     {
       label: 'Call Functions',
       url: buildUrl(`/sandbox/contract-call`, networkMode),
       icon: <Box as={FunctionIcon} size="24px" />,
+      isSelected: router.pathname === '/sandbox/contract-call',
```
leahjlou, line 36: It looks like the selected state isn't working in the UI for the `contract-call` page, but it is for the other two pages

leahjlou: This was just before I had connected my wallet, but it works after I connect the wallet. So not a big deal, but a bit strange

leahjlou: Screen Shot 2022-11-08 at 3 57 08 PM

leahjlou (approved):
This is a huge improvement, nice work 👏  I tested all basic functionality within each section of the Sandbox and it looks like everything's still working, but it would be good to get a few more eyes on it before shipping because it's such a large change

src/modules/sandbox/hooks/useUser.ts
leahjlou: It looks like this is working when I test it, so it's probably fine -- but there might be an edge case I'm not considering in my manual tests

src/modules/sandbox/components/SideNav.tsx
He1DAr: yep it doesn't include query string

src/modules/sandbox/hooks/useUser.ts
He1DAr: The intention of this effect is to check if a user is already logged in (after reloading/revisiting the website) and set user data, so they don't need to log in again after reloading. When a user disconnects, we set the user data to undefined in the reducer.

src/modules/sandbox/components/SideNav.tsx
He1DAr: nice catch! it needs to be `/sandbox/contract-call/[[...params]]`

src/features/search/items/address-result-item.tsx
He1DAr: I wanted to do this initially but thought maybe it's premature optimization as this query is only used here, no strong opinion. I also wanted to refactor query keys across the app as they're now scattered everywhere (some hardcoded, some defined multiple times in different constant files)

PR #915, opened 2022-11-10
feat: remove babel and use swc

### LINKED ISSUE #914: remove babel and use swc
Use Next.js rust compiler (swc) instead of Babel, which is is [17x faster](https://nextjs.org/docs/advanced-features/compiler)

He1DAr:
closes #914


leahjlou (approved):
Ohh, nice!

PR #918, opened 2022-11-11
feat: use historical STX price

### LINKED ISSUE #917: Use historical STX price
For completed transactions, we should surface historical STX $ value.

He1DAr:
closes #917 
## caveats:
Coingecko API accepts time granularity of **days** (dd-mm-yyyy), so we use tx block burn day to get the price instead of the exact moment of triggering the tx.


leahjlou (approved):
looks great!

PR #925, opened 2022-11-22
feat: stacks 2.1

He1DAr:
- display clarity version for smart contract txs
- display alt recipient for coinbase txs
- display memo for STX transfer events

### DIFF
--- a/src/components/transaction-details.tsx
+++ b/src/components/transaction-details.tsx
@@ -288,7 +288,34 @@ const transformDataToRowData = (tx: Transaction | MempoolTransaction, block?: Bl
         },
         children: tx.coinbase_payload.data,
       };
-      return [txid, sender, fees, nonce, blockTime, blockHash, scratch, canonical];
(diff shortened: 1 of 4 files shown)


src/components/transaction-details.tsx
```diff
@@ -288,7 +288,34 @@ const transformDataToRowData = (tx: Transaction | MempoolTransaction, block?: Bl
...
+        label: {
+          children: 'Alt recipient',
+        },
+        children: tx.coinbase_payload.alt_recipient,
+      };
+      return [txid, sender, fees, nonce, blockTime, blockHash, scratch, canonical, altRecipient];
+    }
+    case 'smart_contract': {
+      const clarityVersion = {
+        condition: !!tx.smart_contract?.clarity_version,
```
zone117x, line 302: It would be correct and backward compatible to default to `clarity_version: 1` if the value isn't available.

zone117x (approved):

PR #929, opened 2022-11-29
feat: pass network env vars as props

He1DAr:
- Passing network environment variables as props to overcome the limitation that docker+nextjs have when accessing environment variables from the client, more details [here](https://github.com/vercel/next.js/discussions/17641).
- Eliminating `next-redux-wrapper` due to a bug (https://github.com/kirill-konshin/next-redux-wrapper/issues/496https://github.com/kirill-konshin/next-redux-wrapper/issues/505, https://github.com/kirill-konshin/next-redux-wrapper/issues/496)


src/common/constants.ts
```diff
@@ -67,6 +56,11 @@ export const DEFAULT_TESTNET_SERVER =
   process.env.NEXT_PUBLIC_TESTNET_API_SERVER ||
   'https://stacks-node-api.testnet.stacks.co';
 
+console.log('config?.NEXT_PUBLIC_MAINNET_API_SERVER', config?.NEXT_PUBLIC_MAINNET_API_SERVER);
```
leahjlou, line 59: Did you mean to leave these console.logs in here?

src/common/state/network-slice.ts
```diff
@@ -1,27 +1,48 @@
...
 const RELOAD_DELAY = 500;
 
-const reloadWithNewMode = (networkMode: string) =>
+const reloadWithNewNetwork = (network: Network) =>
   setTimeout(() => {
     if (typeof window !== 'undefined') {
       const href = new URL(window.location.href);
-      href.searchParams.set('chain', networkMode);
+      href.searchParams.set('chain', network.mode);
+      console.log('hrefff network', network);
```
leahjlou, line 38: Stray console.logs

src/common/utils.tsx
```diff
@@ -473,3 +474,11 @@ export function getNextPageParam(options?: { limit: number; offset: number; tota
   if (Math.abs(delta) === sum || isAtEnd) return undefined;
   return sum;
 }
+
+export const getCustomNetworksFromLS = () => {
+  try {
+    return JSON.parse(localStorage.getItem(CustomNetworksLSKey) || '{}') as Record<string, Network>;
+  } catch (_) {
+    return {};
```
leahjlou, line 482: This would only throw if the item saved in LS were invalid JSON, right? In the error case, would we want to clear the LS item?

src/components/links.tsx
```diff
@@ -1,29 +1,38 @@
...
+  if (network?.isCustomNetwork) {
+    return `${url}&api=${network.url}`;
+  }
+  return url;
+};
 
-export const TxLink: React.FC<{ txid: string }> = React.memo(({ txid, ...rest }) => (
-  <Link href={buildUrl(`/txid/${encodeURIComponent(txid)}`)} passHref {...rest} />
-));
+export const StacksLink: FC<{ path: string }> = ({ path, ...rest }) => {
```
leahjlou, line 15: This wrapper is nice 👍 

This is nitpicky -- but the name, at first, made me think it would be a wrapper for a stacks transaction. Do you think `ExplorerLink` might be a better name? (Hopefully code automation tools could make changing this easy 😅)

src/pages/_app.tsx
```diff
@@ -66,38 +65,29 @@ function ExplorerApp({ Component, ...rest }: ExplorerAppProps): React.ReactEleme
...
-      queryNetworkMode,
-      activeNetwork,
-      DEFAULT_NETWORK_MAP,
-      NetworkModeUrlMap
-    );
-    store.dispatch(setActiveNetwork(DEFAULT_NETWORK_MAP[NetworkModeUrlMap[queryNetworkMode]]));
-  }
+  const queryApiUrl = Array.isArray(query.api) ? query.api[0] : query.api;
+  store.dispatch(initialize({ queryNetworkMode, apiUrls: NetworkModeUrlMap, queryApiUrl }));
+  console.log(
```
leahjlou, line 75: Do we still want to keep all these debug logs? Are they useful?

leahjlou:
Nice change, I tested the functionality and can't find any issues.

Question: What is `lockups.json` and why was it removed?

src/common/utils.tsx
He1DAr: I added try/catch primarily to not throw if localStorage is undefined

src/pages/_app.tsx
He1DAr: will only keep them until 2.1 is out to help spot issues quickly

He1DAr:
> Nice change, I tested the functionality and can't find any issues.
> 
> Question: What is `lockups.json` and why was it removed?

It's an old file that has no references in the code.

leahjlou (approved):
Great changes 👏

PR #943, opened 2022-12-15
feat: exclude block info from txs in microblocks

### LINKED ISSUE #885: Misleading transaction time when in micro block
**Describe the bug**
A time in the past is shown for transactions in micro blocks

The time does not makes sense as it looks like that the tx is executed at the already confirmed block, more precisely, that the clarity variable `block_height` is the height of the confirmed block. This is not true. 

**To Reproduce**
1. Go to https://explorer.stacks.co/transactions?chain=testnet
2. See timestamp on microblock

**Expected behavior**
Do not show a transaction time until the block with the block_height used during processing the transaction is confirmed on Bitcoin.

**Screenshots**
image

He1DAr:
closes #885

### DIFF
--- a/src/components/transaction-details.tsx
+++ b/src/components/transaction-details.tsx
@@ -174,7 +174,10 @@ const transformDataToRowData = (tx: Transaction | MempoolTransaction, block?: Bl
     children: <FeesComponent tx={tx} />,
   };
   const blockTime = {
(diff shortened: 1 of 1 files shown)


rafaelcr (approved):
💯

leahjlou (approved):

PR #946, opened 2022-12-16
feat: contract call improvements

He1DAr:
- Support list arguments
- Break down contract-call page into smaller components & migrate it to modules
- Support calling functions from contract page
- Fix imports
image
image
image


src/modules/sandbox/components/Argument/ListArgument.tsx
```diff
@@ -0,0 +1,66 @@
...
+              <Argument
+                name={`${name}.${i}`}
+                type={listItemsType}
+                handleChange={handleChange}
+                error={error}
+                value={listItemValue}
+              />
+              <CloseIcon
+                size={3}
+                ml={'10px'}
```
leahjlou, line 47: nit: any way to use a flex gap on the parent instead of a left margin?

src/modules/sandbox/components/Argument/PrimitiveArgument.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { FC } from 'react';
+
+import { ClarityAbiType, getTypeString } from '@stacks/transactions';
+import { Box, Input, color } from '@stacks/ui';
+
+import { Caption, Text } from '@components/typography';
+
+import { CommonArgumentProps } from './types';
+
+export const PrimitiveArgument: FC<
```
leahjlou, line 10: nit: a name like 'PrimitiveArgumentInput' would make it clearer what this component is

src/modules/sandbox/components/Argument/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { ClarityAbiTypeTuple, isClarityAbiList } from '@stacks/transactions';
+
+import { ListValueType, TupleValueType } from '../../types';
+import { getTuple } from '../../utils';
+import { ListArgument } from './ListArgument';
+import { PrimitiveArgument } from './PrimitiveArgument';
+import { TupleArgument } from './TupleArgument';
+import { ArgumentProps, CommonArgumentProps } from './types';
+
+export const Argument: FC<CommonArgumentProps & ArgumentProps> = ({
```
leahjlou, line 12: This is really nice 💯

PR #960, opened 2023-01-17
Feat: Add post conditions to snadbox function calls

BLuEScioN:
Issue: https://github.com/orgs/hirosystems/projects/18?pane=issue&itemId=6200877

Feature: Adds post conditions to sandbox function view. This markets this feature of Clarity, it allows us to relax the deny mode requirement, and in allow mode it enables the user to add standard post condition safeguards. Adds a toggle for allow deny mode 

UX:
https://user-images.githubusercontent.com/20312835/213023115-1f2a6521-171e-48d4-8b8f-20aac67bff6b.mov

Testing:
NonFungible Post Condition -https://explorer.stacks.co/txid/0x9b77a3f342f1f8227e4395d8ec2649ed5bcb058ecfe44c811489174430e756fc?chain=mainnet


andresgalante:
Thanks @BLuEScioN, this is really good! Working without a designer is challenging, and this interaction is tricky.

Here is my take on custom selects: Unless necessary, we should try to avoid them. In this case, it doesn't matter because we'll change the UI layer next week, but here is why:

1- Accessibility on a select is extremely hard to nail. A native select is accessible by default
2- In general, unless the design is spot on, custom selects perform worst when doing user research.
3- It's more code to maintain, and it can fail or have bugs.

Having said that, I wouldn't change this PR.

BLuEScioN:
> Thanks @BLuEScioN, this is really good! Working without a designer is challenging, and this interaction is tricky.
> 
> Here is my take on custom selects: Unless necessary, we should try to avoid them. In this case, it doesn't matter because we'll change the UI layer next week, but here is why:
> 
> 1- Accessibility on a select is extremely hard to nail. A native select is accessible by default 2- In general, unless the design is spot on, custom selects perform worst when doing user research. 3- It's more code to maintain, and it can fail or have bugs.
> 
> Having said that, I wouldn't change this PR.

I can change the dropdown to use select. Should only require minor css retouching

He1DAr (approved):

juliencarbonnell:
How's it going, guys? 
this allow/deny switch for post-conditions would be of great use!

PR #1055, opened 2023-03-17
fix: dockerfile yarn

He1DAr:


Dockerfile
```diff
@@ -10,6 +10,9 @@ WORKDIR /app
 
 COPY . .
 
+RUN node -v
+RUN yarn -v
```
smcclellan, line 14: Would `npm -v` also be useful here?

Issue #956, opened 2023-01-12
Sandbox fails with optional function parameter 


whoabuddy:
**Describe the bug**

When submitting an optional value from the sandbox the transaction is rejected before broadcast. 

So far I've seen this in two contract functions on testnet:

- `register-user` in [miamicoin-core-v2](https://explorer.stacks.co/address/ST1H1733V5MZ3SZ9XRW9FKYGEZT0JDGEB8WRH7C6H.miamicoin-core-v2?chain=testnet)
- `transfer` in [miamicoin-token-v2](https://explorer.stacks.co/address/ST1H1733V5MZ3SZ9XRW9FKYGEZT0JDGEB8WRH7C6H.miamicoin-token-v2?chain=testnet)

The behavior of the empty field makes me suspect these are not being wrapped in some/none.

**To Reproduce**

1. Go to the Explorer sandbox
2. Enter a contract name to pull up the functions
3. Select a function that accepts an optional value
4. Try to submit TX and see the error

**Expected behavior**

The transaction would succeed with either `(none)` or `(some xyz)` as the value.

**Screenshots**

![Screenshot from 2023-01-12 07-05-25](https://user-images.githubusercontent.com/9038904/212089745-3b4ec743-46be-4ba7-b443-41a0dca49551.png)
![Screenshot from 2023-01-12 07-05-54](https://user-images.githubusercontent.com/9038904/212089746-48e5b338-ec36-4751-b710-85fc5c73fcca.png)
![Screenshot from 2023-01-12 07-06-16](https://user-images.githubusercontent.com/9038904/212089748-dfd02eaa-867f-486c-bc26-b46bba23114a.png)
![Screenshot from 2023-01-12 07-06-27](https://user-images.githubusercontent.com/9038904/212089750-9bba15c6-2364-4a8e-9dc2-8ace978aedc0.png)

314159265359879:
I have noticed this too. Testing for ALEX. 

The workaround I used was to deploy a contract instead with the contract call in it.

janniks:
Yeah +1
Putting this on the backlog. The explorer currently encodes `optional(string)` as `string` regardless of if the string is empty or has value (maybe if it's empty it should default to `optional-none`

janniks:
I don't think it's a stacks.js / wallet bug -- this example app using https://stacks.new seems to work: https://codesandbox.io/p/sandbox/cool-paper-fesyxn?welcome=true

andresgalante:
[...]

PR #1066, opened 2023-03-22
fix(sandbox): handle optional function args (#956)

friedger:
This PR
* handles non-tuple optional function arguments in the sandbox

This change is important for using city coin contracts, wrapper pox contracts, etc.


janniks:
Got it, thanks -- Nice change 👏 Will defer to explorer team for a review/merge

> **Note**
>  @ explorer Originally, I think the `encodeClarityValue` is technically isn't meant for what the explorer is doing here, (and doesn't implement all types) -- that was the original problem, so we might run into the same thing again for other wrapper/composite types.

He1DAr (approved):

PR #1073, opened 2023-03-27
fix: tests on CI

He1DAr:


src/common/utils/token-utils.spec.ts
```diff
@@ -3,6 +3,7 @@ import { isIconUrl } from './url-utils';
 
 describe(isFtNameLikeStx.name, () => {
   it('detect impersonating token names', () => {
+    expect(isFtNameLikeStx('TEST_FAIL')).toBeTruthy();
```
leahjlou, line 6: Shouldn't this be `toBeFalsy`?

PR #1099, opened 2023-04-20
feat: hide pox-2.stack-increase

friedger:
This PR
* hides pox-2.stacks-increase calls in preparation for Stacks 2.2
  * in the sandbox's list of available functions
  * in the contract page's list of available functions
  * in the sandbox deep link http://localhost:3000/sandbox/contract-call/SP000000000000000000002Q6VF78.pox-2/stack-increase?chain=mainnet
  
image

How to test?
Check the list of available functions on the following pages:
* visit /txid/SP000000000000000000002Q6VF78.pox-2?chain=mainnet 
* visit sandbox/contract-call/SP000000000000000000002Q6VF78.pox-2?chain=mainnet
* visit sandbox/contract-call/SP000000000000000000002Q6VF78.pox-2/stack-increase?chain=mainnet

### DIFF
--- a/src/app/common/utils/sandbox.ts
+++ b/src/app/common/utils/sandbox.ts
@@ -1,0 +1,8 @@
+export function showFn(contractId: string, abiFn: any) {
+  return (
+    abiFn.access !== 'private' &&
+    (abiFn.name !== 'stack-increase' ||
+      (contractId !== 'SP000000000000000000002Q6VF78.pox-2' &&
+        contractId !== 'ST000000000000000000002AMW42H.pox-2'))
+  );
+}
--- a/src/app/common/utils/sandbox.ts
+++ b/src/app/common/utils/sandbox.ts
@@ -1,3 +1,13 @@
+/**
+ * Predicate that a function of a contract should be shown.
+ *
+ * Functions that are private should NOT be shown
+ * Functions that are pox-2.stacks-increase should NOT be shown
+ *
+ * @param contractId the contract id
+ * @param abiFn the abi of the function or
+ * @returns true if the provided function should be shown in the explorer.
+ */
 export function showFn(contractId: string, abiFn: any) {
   return (
     abiFn.access !== 'private' &&
--- a/src/app/common/utils/sandbox.ts
+++ b/src/app/common/utils/sandbox.ts
@@ -2,7 +2,8 @@
  * Predicate that a function of a contract should be shown.
  *
  * Functions that are private should NOT be shown
- * Functions that are pox-2.stacks-increase should NOT be shown
+ * Functions that are pox-2.stacks-increase should NOT be
+ * shown as it is invalid since Stacks 2.2
  *
  * @param contractId the contract id
  * @param abiFn the abi of the function or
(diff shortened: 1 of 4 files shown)


src/app/sandbox/components/ContractCall/AvailableFunctionsView.tsx
```diff
@@ -67,14 +67,23 @@ export const AbiFunction = forwardRef<
...
+
 export const AvailableFunctionsView: FC<{
   contract: ContractWithParsedAbi;
   contractId: string;
 }> = ({ contractId, contract }) => (
   <Section overflowY="auto" flexGrow={1} title="Available functions">
     {contract?.abi?.functions.map(
       (abiFn: any) =>
-        abiFn.access !== 'private' && (
+        hideFn(contractId, abiFn) && (
```
diwakergupta, line 86: did you mean `showFn`?

friedger: fixed

src/app/common/utils/sandbox.ts
```diff
@@ -0,0 +1,8 @@
+export function showFn(contractId: string, abiFn: any) {
+  return (
```
diwakergupta, line 2: Can you add a brief comment to document this function and in particular, make a note about stacks-increase in PoX-2.

Separately, I found this implementation a bit hard to reason through: basically we're saying show a function IF:
* it's NOT private, _and_
* it NOT stacks-increase in POX-2

IMO an implementation that uses fewer negatives might be easier to understand and maintain. Something like:

hide a function IF
* it's private OR
* it's stacks-increase in testent/mainnet pox-2

src/app/sandbox/components/ContractCall/AvailableFunctionsView.tsx
```diff
@@ -10,7 +11,7 @@ import { Text } from '@/ui/typography';
 import { forwardRef, useColorMode } from '@chakra-ui/react';
 import { mdiApi, mdiFunction } from '@mdi/js';
 import Icon from '@mdi/react';
-import React, { FC } from 'react';
```
diwakergupta, line 13: This seems like an unrelated removal? Can we avoid any unnecessary changes in the PR to simplify the diff? Thanks.

diwakergupta (changes requested):
I'm not super familiar with the explorer code base but overall changes LGTM. If you can add some docs to `showFn` we can get this merged. Thanks Friedger!

src/app/sandbox/components/ContractCall/AvailableFunctionsView.tsx
friedger: this happened when I ran format

src/app/common/utils/sandbox.ts
friedger: My motivation was to have a cleaner code in the components:
`showFn(contractId, abiFn) && <FunctionView.../>`

and when using filter the abi

diwakergupta (approved):
LGTM!

Issue #1100, opened 2023-04-21
There should be proper error handling gracefully using try catch


shamim-io:
**Describe the bug**
There should be proper error handling gracefully using try catch for api calls in useEffect

**To Reproduce**
1. Go to homepage but turning off the internet connectivity
2. See error: The page breaks as the api call doesn't take page successfully. see image below.

**Expected behavior**
The page should not break and error message should be there justifying the error.

**Screenshots**
<img width="1415" alt="Screenshot 2023-04-21 at 7 25 10 PM" src="https://user-images.githubusercontent.com/29030948/233663815-28c550cd-fa03-4db9-aee9-bc6e6e25b9de.png">

shamim-io:
Can I take up the issue, solve it and raise a PR?

andresgalante:
Yes please!  That'd be great, PRs are very welcome!

shamim-io:
The branch has been merged. Can we close the issue?

andresgalante:
Thanks for the contribution @shamim-io !

PR #1107, opened 2023-04-30
refactor: handled api calls gracefully

shamim-io:


shamim-io:
solves [There should be proper error handling gracefully using try catch #1100 ]

Issue #1108, opened 2023-04-30
Component app-config.tsx can be optimized


shamim-io:
**Describe the bug**

Transaction ID 
Address 
Block#  
Time stamp

Component `app-config.tsx` can have following enhancements: 
1. Define interface for the component props to avoid using 'any' type.
2. Memoize the handleRouteChangeComplete function using useCallback to prevent unnecessary re-renders.
3. Add dependencies to the useEffect hooks to prevent unnecessary re-renders.
4. Remove console.log statements.
5. Improve the code readability by properly formatting the JSX code.

shamim-io:
Can I take up the issue? @andresgalante

andresgalante:
Yes, please!! a PR is very welcome! 🙏

shamim-io:
@andresgalante I have raised a PR. Please check it and let me know if everything is okay. Thanks!

andresgalante:
Thanks for the PR!  🎉 We'll review it today

andresgalante:
I know this PR was closed. If you think you have a case for it and numbers to justify the performance improvement, we are happy to reconsider it.

Else please feel free to close the issue 🙏

PR #1111, opened 2023-05-01
refactor: app-config

shamim-io:
1. Used useCallback to memoize the events.on function in the useEffect Hook to avoid redefining functions and reduce unnecessary re-renders of components.
2. Used useMemo to memoize the toast message in the useEffect Hook to memoize values between renders and recompute it only if the dependencies change.
3. Used useMemo to memoize the appDetails object in the Connect component to avoid creating new instances of the object on every render.
4. Removed unnecessary console.logs to improve performance.
5. Made sure to include all dependencies in the useCallback and useMemo Hooks to ensure that the memoized value is updated only when necessary.

Solves: Component app-config.tsx can be optimized #1108

### DIFF
--- a/src/components/app-config.tsx
+++ b/src/components/app-config.tsx
@@ -5,7 +5,7 @@ import { NetworkMode } from '@/common/types/network';
 import { cache } from '@emotion/css';
 import { CacheProvider } from '@emotion/react';
 import { useRouter } from 'next/router';
-import React, { useEffect } from 'react';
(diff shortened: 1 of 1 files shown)


He1DAr:
In these instances, the returned value is pretty simple where using it directly is likely to be more performant compared to the overhead of using memorization. I'm happy to review again if you can quantify the performance improvement. Thank you!

PR #1112, opened 2023-05-03
Updates the banner with information about SIP 23

andresgalante:


saralab (approved):
LGTM , let's ship it!

PR #1134, opened 2023-05-12
Fix banner

friedger:
This PR
* updates the banner to say that sip 22 explains about the two hard forks and that currently, stacking is disabled


friedger:
I still meet people that don't know that stacking is disabled right now.

He1DAr:
updated the alert in a separate pr https://github.com/hirosystems/explorer/pull/1135

Issue #1154, opened 2023-05-30
fetcher function in fetch.ts is not handled gracefully.


shamim-io:
**Describe the bug**
The fetcher function in the `src/common/api/fetch.ts` is not handled gracefully. That's the reason it's not throwing appropriate error. 

Currently it shows `TypeError` whenever an api call fails. 

**Expected behavior**
It should show it as `Network Error`

**Screenshots**
<img width="694" alt="Screenshot 2023-05-30 at 11 24 55 PM" src="https://github.com/hirosystems/explorer/assets/29030948/cf4882ee-d7b6-4566-8a15-d23e67b7b67b">

PR #1155, opened 2023-05-30
fix: handled fetcher function gracefully

shamim-io:
closes https://github.com/hirosystems/explorer/issues/1154

Handled fetcher function gracefully

Previous code: 
Screenshot 2023-05-30 at 11 24 55 PM

Modified code: 
Screenshot 2023-05-30 at 11 35 51 PM

Files changed
`src/common/api/fetch.ts`

### DIFF
--- a/src/common/api/fetch.ts
+++ b/src/common/api/fetch.ts
@@ -15,8 +15,8 @@ export const fetcher = async (input: RequestInfo, init: RequestInit = {}) => {
 
 export const fetchFromApi =
   (apiServer: string) =>
-    async (path: string, opts = {}) => {
-      return fetcher(withApiServer(apiServer)(path), {
-        ...opts,
-      });
-    };
+  async (path: string, opts = {}) => {
+    return fetcher(withApiServer(apiServer)(path), {
+      ...opts,
+    });
+  };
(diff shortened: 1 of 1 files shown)


He1DAr (approved):

src/common/api/fetch.ts
```diff
@@ -1,12 +1,16 @@
...
-    headers: { ...initHeaders, ...HIRO_HEADERS },
-  });
+  try {
+    return await fetch(input, {
+      credentials: 'omit',
+      ...init,
+      headers: { ...initHeaders, ...HIRO_HEADERS },
+    });
+  } catch (error) {
+    throw new Error(`Network error: ${error.message}`);
```
He1DAr, line 12: Hi @shamim-io! Thank you for submitting this PR. Could you explain the motivation behind this change? It looks like the error is being rethrown.

He1DAr (changes requested):

src/common/api/fetch.ts
shamim-io: @He1DAr If you think that, its not a required change. Feel free to close it :)

He1DAr:
Thank you for submitting this PR!
I think the new implementation will behave exactly the same, except it adds "Network error" to the error message when it throws. Here's an example:
image
image
Since we want to keep throwing the error, I suggest we close the PR. Please let me know your thoughts on this.

shamim-io:
I agree with you. Thanks @He1DAr for the clarification. Closing the PR!

PR #1194, opened 2023-06-23
fix: make action buttons from `TxListTabs` semantic and aligned

### LINKED ISSUE #1188: The Export CSV btn is out of place
The `Export as CSV` button on the address page is out of place. It should be a sibling of the filter action and be placed at the left of the filters. (For example https://explorer.hiro.so/address/SPQR489BFKKAQPNF3KRTAVVBJ404DWY42KRPHQQ0?chain=mainnet)

Screenshot 2023-06-21 at 20 55 09

Bonus points if since we are at it, we convert it from a span to a `<button>` so the HTML is semantic.

13Five5:
Fixes #1188

### DIFF
--- a/src/components/CSVDownloadButton.tsx
+++ b/src/components/CSVDownloadButton.tsx
@@ -2,7 +2,6 @@
 
 import { Box, Icon } from '@/ui/components';
 import { CSVDownload } from 'react-csv';
-import { Caption } from '@/ui/typography';
 import React, { memo, useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


andresgalante:
@raphaeljp  Thanks for the contribution! there is only one more step for us to review it: We need you to [sign the CLA](https://github.com/hirosystems/explorer/pull/1194#issuecomment-1603619660) in order to merge the PR.

Would you be able to do it please?

Thanks!

Issue #566, opened 2021-10-18
Use brace syntax for tuples, instead of (tuple ...) when displaying events


lgalabru:
![Screen Shot 2021-10-18 at 5 12 34 PM](https://user-images.githubusercontent.com/87777/137807424-64660ea4-d70c-41b2-8ead-b8d0765e6291.png)

Using a `{ action: "list-item", data: { ... }}` syntax would make the UI clearer.

PR #1236, opened 2023-07-14
Fix: display data in json format

dragnoir:
Close [#566](https://github.com/hirosystems/explorer/issues/566)

I didn't remove the "(tuple " at the start of the data because in some other cases, it's something else and there's more "tuple" in a deeper level of the data.

before:
image

after:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.110.2](https://github.com/hirosystems/explorer/compare/v1.110.1...v1.110.2) (2023-07-05)
+
+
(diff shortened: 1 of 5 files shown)


He1DAr:
@andresgalante @chresko Could you please take a look at this change? My concern is it can give a false impression that return value is JSON.

andresgalante:
@dragnoir, great work! This is definitely way more clear to read.

I share the same concern as @He1DAr , would you be able to add limiting and indentation to the original format without converting the output instead of making it json like?

PR #1249, opened 2023-07-30
fix: holdings indicator incosistent

### LINKED ISSUE #1235: Holdings indicator in the address page is incosistent
The holdings indicator in the summary panel is inconsistent with the holdings panel on the right

Here is an example:
https://explorer.hiro.so/address/SPTW196D2867VBY2CAH71G8MMGJ7QYDMT7MZPG7Q?chain=mainnet

It indicates that there are 2 tokens, but the holdings panel shows none.

image

bharat2913:
This PR closes #1235 

Holdings indicator in both left and right panels are consistent now.
image


He1DAr (approved):

andresgalante:
@He1DAr Are we ready to ship this one?

PR #1263, opened 2023-08-07
feat: add custom network using query param

### LINKED ISSUE #1257: Adding a custom Stacks network does not work with full URL path
**Describe the bug**
When adding a custom network through the "Network" menu, when using a URL with a path (i.e. `http://localhost:8477/api/v1/network/my-network/stacks-api`), the network cannot be added because the wrong `/info` API request is made. The error message displayed in the UI is: `JSON.parse: unexpected character at line 1 column 1 of the JSON data`.

If you open the devtools Network tab and view the request made to check the API info, you can see that the request is made to `GET http://localhost:8477/v2/info`. The request should instead include the full path: `GET http://localhost:8477/api/v1/network/my-network/stacks-api/v2/info`.

**To Reproduce**
1. Go to any page in the Explorer
2. Click the "Network" menu and click "Add a Network"
3. Add a network with the API URL `http://localhost:8477/api/v1/network/my-network/stacks-api`
4. See that the network cannot be added.

**Expected behavior**
The request to `/v2/info` should be made including the full URL path, and the network should be added if it is available.

**Screenshots**
Screen Shot 2023-08-04 at 12 18 09 PM

leahjlou:
When opening the Stacks Explorer with an API URL in the query string (`api`), if that network does not exist and it is a valid Stacks API URL, add it to the list of custom networks. This allows an external application to link to the Explorer with a prefilled custom Stacks network, and the Explorer will automatically open to that network.

Also, support custom networks whose Base URL includes a path (fixes #1257).

### DIFF
--- a/src/common/context/GlobalContext.tsx
+++ b/src/common/context/GlobalContext.tsx
@@ -1,11 +1,12 @@
 import { DEFAULT_DEVNET_SERVER, IS_BROWSER } from '@/common/constants';
-import { NetworkModeUrlMap } from '@/common/constants/network';
+import { NetworkIdModeMap, NetworkModeUrlMap } from '@/common/constants/network';
 import { Network, NetworkModes } from '@/common/types/network';
 import cookie from 'cookie';
-import { FC, createContext, useMemo, useState } from 'react';
+import { FC, createContext, useCallback, useEffect, useMemo, useState } from 'react';
 import { useCookies } from 'react-cookie';
 
 import { ChainID } from '@stacks/transactions';
+import { buildCustomNetworkUrl, fetchCustomNetworkId } from '@/components/add-network-form';
 
 interface GlobalContextProps {
   cookies: string;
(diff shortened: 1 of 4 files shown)


src/components/add-network-form.tsx
```diff
@@ -22,16 +22,20 @@ import { string } from 'yup';
 
 import { ChainID } from '@stacks/transactions';
 
-const buildCustomNetworkUrl = (url: string) => {
-  const hostname = encodeURIComponent(new URL(url).hostname);
-  const port = encodeURIComponent(new URL(url).port);
-  return `${hostname === 'localhost' ? 'http://' : 'https://'}${hostname}${port ? `:${port}` : ''}`;
+export const buildCustomNetworkUrl = (url: string) => {
```
He1DAr, line 25: It would be great if you add a unit test to this function.

src/common/context/GlobalContext.tsx
```diff
@@ -84,33 +85,57 @@ export const AppContextProvider: FC<any> = ({
     [apiUrls, customNetworks, isUrlPassedSubnet, querySubnet]
   );
 
+  const addCustomNetwork = useCallback(
```
He1DAr, line 88: And this too if possible.

He1DAr (approved):

src/common/context/__tests__/GlobalContext.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+const apiUrls = {
+  mainnet: 'https://api.hiro.so',
+  testnet: 'https://api.testnet.hiro.so',
+};
+
+const getContextField = (fieldId: string) => {
+  return JSON.parse(screen.getByTestId(fieldId).innerHTML);
+};
+
+describe('GlobalContext', () => {
```
leahjlou, line 31: Added this test just for the new behavior. It can be expanded later to test more of the global context functionality.

PR #1301, opened 2023-10-04
Stacks Explorer Sandbox - feature guide

elenananana:
I plan to come back to this, just wanted to submit these changes in case I lose them. #githubnoob


elenananana:
@He1DAr - here's a feature guide for sandbox explorer. LMK what you think. The only outstanding item here is
to write an example for calling an NFT mint function from the Sandbox, I would just need a mint contract and I think I can take it from there. thanks!

andresgalante (approved):

PR #1324, opened 2023-10-31
specified pnpm version

niwim:
Suggested fix for issue 1323 - Docker build failing.

The build is breaking because of a new release of pnpm being incompatible. This should probably be fixed by understanding why the new release is breaking, but this should work for now.


He1DAr:
Thank you @niwim!

PR #1362, opened 2023-12-16
Feat/browser variables docker

niwim:
Browser variables are set at build time through NEXT_PUBLIC variables -https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

As the build arguments are currently not exposed in the Dockerfile the only way to set these variables during build is through the .env.local file.

Adding the NEXT_PUBLIC as build arguments in the Dockerfiles makes it possible to use environment variables instead and makes for a flexible solution when building a pipeline.


niwim:
Sorry, screwed up the rebase, will make new PR

PR #1431, opened 2024-01-18
feat: block list layout A

He1DAr:
image


src/app/PageClient.tsx
```diff
@@ -30,7 +38,12 @@ export default function Home() {
         gridTemplateColumns={['100%', '100%', 'minmax(0, 0.6fr) minmax(0, 0.4fr)']}
       >
         <TxListTabs limit={DEFAULT_LIST_LIMIT_SMALL} />
-        <BlocksList limit={DEFAULT_BLOCKS_LIST_LIMIT} />
+
+        {activeNetworkKey.indexOf('naka') !== -1 ? (
```
BLuEScioN, line 42: Should make this a function

src/app/_components/BlockList/LayoutA/BlockCount.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { PiArrowUpRight } from 'react-icons/pi';
+
+import { Circle } from '../../../../common/components/Circle';
+import { ExplorerLink } from '../../../../common/components/ExplorerLinks';
+import { Flex } from '../../../../ui/Flex';
+import { Icon } from '../../../../ui/Icon';
+import { Text } from '../../../../ui/Text';
+
+export const BlockCount = memo(function ({ count }: { count: number }) {
+  const bgColor = useColorModeValue('purple.100', 'slate.900');
```
BLuEScioN, line 13: I'm seeing a lot of this: defining the color with hooks. Not a huge fan of this as it creates a lot of noise that gets mixed in with the functional logic. Not suggesting we do things differently rn but I think it would be worth it at some point to find a different solution that seperates this logic out of the component would be more manageable.

BLuEScioN (approved):

src/app/_components/BlockList/LayoutA/BlockCount.tsx
He1DAr: I agree, for the default color configuration, we add them to the theme. However, these are specific to the component. LMK if you have any suggestions. Maybe we could move them to a hook that is collocated with each component?

src/app/PageClient.tsx
He1DAr: Makes sense. It's a temporary way to display the new list for testing, once Nakamoto is out, we'll entirely remove the old list and the forking logic.

PR #1434, opened 2024-01-19
fix: hello world contract

hugocaillard:
Missing a `)`


He1DAr (approved):

He1DAr:
Thank you @hugocaillard! Can you please fix the unit test so we can ship it?

hugocaillard:
@He1DAr I see that the linting is broken on multiple PRs here, can it be fixed on main instead?

Issue #1413, opened 2024-01-09
do not show error message for a token that is not SIP-10 compliant


rafaelcr:
<img width="350" alt="Screenshot 2024-01-09 at 9 05 29" src="https://github.com/hirosystems/explorer/assets/216195/a12a7ab4-498a-47aa-a2e2-4ac0687c2d89">
The explorer should not display "Unknown error" if the token is not returned by the Metadata API, as these tokens are most likely not SIP-10 compliant.

It should only show users the number of tokens they own while failing silently.

rafaelcr:
Link shown above is: https://explorer.hiro.so/address/SP1K1A1PMGW2ZJCNF46NWZWHG8TS1D23EGH1KNK60?chain=mainnet

andresgalante:
Good point, thanks @rafaelcr

PR #1435, opened 2024-01-19
fix(sip-10): do not show error message for a token that is not SIP-10 compliant

BLuEScioN:
Captured error and no longer show unknown error for non sip 10 fts

Resolves https://github.com/hirosystems/explorer/issues/1413 

image

### DIFF
--- a/src/common/context/GlobalContext.tsx
+++ b/src/common/context/GlobalContext.tsx
@@ -46,7 +46,7 @@ export const GlobalContext = createContext<GlobalContextProps>({
     networkId: ChainID.Mainnet,
     mode: NetworkModes.Mainnet,
   },
-  activeNetworkKey: NetworkModeUrlMap[NetworkModes.Mainnet],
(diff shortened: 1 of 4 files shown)


src/common/context/GlobalContext.tsx
```diff
@@ -46,7 +46,7 @@ export const GlobalContext = createContext<GlobalContextProps>({
     networkId: ChainID.Mainnet,
     mode: NetworkModes.Mainnet,
   },
-  activeNetworkKey: NetworkModeUrlMap[NetworkModes.Mainnet],
+  activeNetworkKey: NetworkModeUrlMap[NetworkModes.Mainnet], // TODO: this is a confusing name as it's actually the url for the api based on the network...
```
He1DAr, line 49: Just to provide some context, we chose this naming convention because we're using the API URL as the unique identifier (key) for the network. If you feel the current name is a bit confusing, please feel free to change it.

andresgalante:
@BLuEScioN feel free to merge and ship this when you are ready.

src/common/context/GlobalContext.tsx
BLuEScioN: cool. I will update in another PR

PR #1443, opened 2024-01-30
feat: block page layout A

He1DAr:
image

image


src/app/_components/BlockList/LayoutA/Blocks.tsx
```diff
@@ -0,0 +1,56 @@
...
+          case UIBlockType.Block:
+            return (
+              <StxBlock
+                key={block.hash}
+                hash={block.hash}
+                height={block.height}
+                timestamp={block.timestamp}
+                txsCount={block.txsCount}
+                icon={
+                  i === 0 || (i > 0 && blockList[i - 1].type === UIBlockType.BurnBlock) ? (
```
BLuEScioN, line 35: Would we be able to succintly define the condition in English. This is hard to decipher

src/app/_components/BlockList/LayoutA/useBlockListWebSocket.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Block } from '@stacks/stacks-blockchain-api-types';
+
+import { UIBlockType, UISingleBlock } from '../types';
+import { useSubscribeBlocks } from '../useSubscribeBlocks';
+
+export function useBlockListWebSocket(
+  initialBlockHashes: Set<string>,
+  initialBurnBlockHashes: Set<string>
+) {
+  const [latestBlocks, setLatestBlocks] = useState<UISingleBlock[]>([]);
```
BLuEScioN, line 13: Why do we need to use state and refs?

src/app/_components/BlockList/LayoutA/useBlockList.ts
```diff
@@ -61,144 +62,93 @@ const createUIBlockList = (
...
-      clearLatestBlocks();
-      return Promise.all([
+  const { latestBlock, latestBlocksCount, clearLatestBlocks } = useBlockListWebSocket(
+    initialBlockHashes,
+    initialBurnBlockHashes
+  );
+
+  const updateList = useCallback(
+    async function () {
+      setFadeEffect(true);
```
BLuEScioN, line 97: Couldnt this just be a setIsUpdateListLoading? and then use that to trigger the fade effect?

src/app/_components/BlockList/LayoutA/useBlockList.ts
```diff
@@ -61,144 +62,93 @@ const createUIBlockList = (
...
-  if (liveUpdates) {
-    if (latestBlock && !displayedBlocks[latestBlock.hash]) {
+    if (liveUpdatesToggled) {
+      setFadeEffect(true);
+      clearLatestBlocks();
+      updateList().then(() => {
+        setFadeEffect(false);
+      });
+    } else if (receivedLatestBlockWhileLiveUpdates && latestBlock) {
       if (latestBlock.burn_block_height === lastBurnBlock.burn_block_height) {
```
BLuEScioN, line 126: This seems like this is handling a blockchain specific issue, like a reorg. If so, it would be helpful to add a comment about that

src/app/_components/BlockList/useSubscribeBlocks.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { useStacksWebSocketClient } from './useStacksWebSocketClient';
+
+interface Subscription {
+  unsubscribe(): Promise<void>;
+}
+
+export function useSubscribeBlocks(handleBlock: (block: NakamotoBlock) => any) {
+  const [subscription, setSubscription] = useState<Subscription>();
+  const stacksWebSocketClient = useStacksWebSocketClient();
```
BLuEScioN, line 14: This could be an issue if the socket hook is used multiple times. In that case, we would want to use a context wrapper to hold on to the socket connection and keep adding subs to it

src/app/_components/BlockList/LayoutA/__tests__/usePaginatedBlockList.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+        result.current.updateList();
+      });
+    });
+
+    waitFor(() => {
+      expect(resetQuerySpy).toHaveBeenCalledWith({ queryKey: ['blockListInfinite'] });
+    });
+  });
+
+  it('should fetch next page when fetchNextPage is called', async () => {
```
BLuEScioN, line 60: isnt this just testing react query's fetch next page?

src/app/_components/BlockList/LayoutA/__tests__/useBlockList.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    latestUIBlocks: [],
+    latestBlocksCount: 0,
+    clearLatestBlocks: jest.fn(),
+    latestBlock: undefined,
+  })),
+}));
+
+jest.mock('../useInitialBlockList', () => ({
+  useInitialBlockList: jest.fn().mockImplementation(() => ({
+    lastBurnBlock: {
```
BLuEScioN, line 20: This is just a generic question. I'm not quite understanding why we need initial and second to last data. Could you explain it to me?

src/common/debug.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const usePrevious = (value: any, initialValue: any) => {
+  const ref = useRef(initialValue);
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
+
+const useEffectDebugger = (effectHook: any, dependencies: any, dependencyNames: any = []) => {
```
BLuEScioN, line 11: where are these being used?

src/app/_components/BlockList/LayoutA/Paginated.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const showLatestBlocksWithFadeEffect = useCallback(() => {
+    setFadeEffect(true);
+    setTimeout(() => {
+      showLatestBlocks();
+      setFadeEffect(false);
+    }, FADE_DURATION);
+  }, [setFadeEffect, showLatestBlocks]);
+
+  const prevLiveUpdatesRef = useRef(liveUpdates);
```
BLuEScioN, line 55: You have a usePrevious hook that could be used here

BLuEScioN (changes requested):
Nice job! I made a few notes. Take a look and I'll review this again when you're ready

src/app/_components/BlockList/LayoutA/useBlockListWebSocket.ts
He1DAr: Refs are used to keep track of values that don't trigger UI updates when they change, unlike state changes, which do cause UI updates.

src/app/_components/BlockList/LayoutA/useBlockList.ts
He1DAr: Could you elaborate more on that?

src/common/debug.ts
He1DAr: I used them to debug issues around creating multiple WS subscriptions, happy to remove them if you think they're not useful or confusing.

src/app/_components/BlockList/useSubscribeBlocks.ts
He1DAr: Thanks for the suggestion. I added memoization and [unit test](https://github.com/hirosystems/explorer/blob/36ff9b2915221a9c4f444c0e23e690fea1d5a377/src/app/_components/BlockList/__tests__/useStacksWebSocketClient.test.tsx) to handle this issue.

src/app/_components/BlockList/LayoutA/useBlockList.ts
BLuEScioN: The name setFadeEffect seems too specific. What you are actually tracking is if the update list operation is loading. Why not then change setFadeEffect to setIsUpdateListLoading as that would communicate a lot more information about why there is a fade effect. Additionally, in the future, if I needed to use this state to trigger a different operation, triggering it off of teh state of the fade effect wouldn't make any semantic sense

src/app/_components/BlockList/LayoutA/__tests__/useBlockList.test.tsx
BLuEScioN: Interesting. ok

src/app/_components/BlockList/useSubscribeBlocks.ts
BLuEScioN: I don't think that unit test is testing the the hook being used in different components

src/common/debug.ts
BLuEScioN: we can keep them. Just curious

src/app/_components/BlockList/LayoutA/Paginated.tsx
BLuEScioN: This still hasn't been addresssed

BLuEScioN (changes requested):
I responded back on some of the same comments that I made before. Could you please take a look again

src/app/_components/BlockList/LayoutA/Paginated.tsx
He1DAr: are you referring to usePrevious in debug.ts?

BLuEScioN: yes

src/app/_components/BlockList/useSubscribeBlocks.ts
BLuEScioN: Just to be clear we don't need to change anything. This is fine for now, but I think we have to be really careful abou thow we use this hook while we don't have a way of maintaining a reference to the socket across components

src/app/_components/BlockList/LayoutA/Paginated.tsx
He1DAr: I feel using a function from debug in actual logic can cause confusion, I'm planning to remove debug.ts for now and we can discuss if we need to bring it back in a better way.

BLuEScioN (approved):

PR #1457, opened 2024-02-20
fix: icon regression

He1DAr:


src/app/_components/BlockList/LayoutA/__tests__/__snapshots__/BlockListWithControls.test.tsx.snap
```diff
@@ -209,7 +209,7 @@ exports[`BlockListWithControls renders correctly 1`] = `
                   <span
                     class="chakra-text css-nvxyq8"
                   >
-                    in 54071 years
+                    in 54085 years
```
BLuEScioN, line 212: Can we automate this such that we don't have to keep updating the tests

src/ui/theme/theme.ts
```diff
@@ -102,8 +74,7 @@ export const theme = extendTheme({
     heading: openSauce.style.fontFamily,
   },
   sizes: {
-    // 4.5: '1.125rem',
```
BLuEScioN, line 105: These should be exactly the same

src/app/_components/BlockList/LayoutA/__tests__/__snapshots__/BlockListWithControls.test.tsx.snap
He1DAr: I'd do this later as we need to fix the regression ASAP.

src/ui/theme/theme.ts
He1DAr: not sure I fully understood

BLuEScioN (commented):
To me this looks more like another issue with the way we are setting size than with the scale
See Icon.tsx where the size sets size but then also height and width and minwidth and minheight
'use client';

import { Icon as CUIIcon, IconProps as CUIIconProps, forwardRef } from '@chakra-ui/react';

import { UIComponent } from './types';

export type IconProps = CUIIconProps & UIComponent;
export const Icon = forwardRef<IconProps, 'div'>(({ children, size, ...rest }, ref) => (
  <CUIIcon
    ref={ref}
    size={size}
    width={size || rest.width}
    height={size || rest.height}
    minWidth={size || rest.minWidth}
    minHeight={size || rest.minHeight}
    {...rest}
  >
    {children}
  </CUIIcon>
));

src/ui/theme/theme.ts
BLuEScioN: Divide 1.125 by 1.125 that becomes 1 rem. Divide 4.5 by 1.125 and that becomes 4

BLuEScioN: Perhaps, this was working before because 4.5 was getting exactly translated into 1.125 but if the goal is to have a scale for mapping pure numbers to rem then I think some of our components that set the size and also height and width and minwidth and minheight might be messing up. I ran into this when fixing our status bubbles in my ui updates PR

andresgalante (approved):
Let's merge this to fix the regression and we can discuss a better implementation later

PR #1462, opened 2024-02-22
fix: favicon

BLuEScioN:
Ok. I am using the android-chrome-512x512.png for the highest clarity

This is what it looks like:
image

There were a lot of favicons to choose from:
image


He1DAr (approved):
LGTM! Would you mind using the existing branch naming convention for future PRs?
```
feat/kebab-case
fix/kebab-case
...
```

PR #1513, opened 2024-03-12
feat: throttle live updates toggle

He1DAr:


andresgalante (approved):
If someone is abusing, the buffer is really annoying :) so yes, lets ship it

PR #1526, opened 2024-03-22
feat(nakamoto-switch): if on testnet, nakamoto is enabled. all switch…

BLuEScioN:
If the network is on testnet, nakamoto checks will pass and enable nakamoto features.
Changed all nakamoto checks to use the useIsNakamoto hook so all the nakamoto check code is written in one place


He1DAr (approved):
LGTM

PR #1602, opened 2024-04-30
fix: bundle size and dependencies

He1DAr:
The purpose of the PR is to reduce bundle size through refactoring 3rd party dependencies:
**- Consolidate icon libraries**
Previously, explorer used a combination of react-icons libraries as well as mdi.js:
image
image
image
Removing all icon libraries and exclusively use @phosphor-icons/react:
image
**- Remove unnecessary dependencies**
image


BLuEScioN (approved):
LGTM

PR #1676, opened 2024-06-14
feat: status bar revamp

### LINKED ISSUE #1159: Implement CMS for alerts
We should follow through and ship this https://github.com/hirosystems/explorer/pull/1149

He1DAr:
- Add CMS support for dynamic status bar updates
- Revamp status bar design according to latest Figma
- Fix multiple status bars layout
- Add dark mode support

-fix nft link in address page

closes #1159, #1695


eugeniadigon:
There are a couple of differencies between the build and the [design](https://www.figma.com/design/31q1g27k69z1yXb6r6vrRD/Stacks-Explorer-Working-File?m=dev&node-id=7429-39393&t=vZ2W0HDMeZV6uFG0-1):

- [ ] Dark mode should have white background banners, and light mode should have dark background banners. This is to create more visual contrast between the UI elements and the banners
- [ ] Links should be coloured
- [ ] For the top banner, text should be centered on the page
- [ ] There should be 3 types of banners: `(i) Info or notice (blue or purple)`, `/!\ Medium warning (orange)`, `/!\ Severe warning (red)`
- [ ] There should be 2 hierarchies of banners: Primary (top of page) and secondary (on hero section). Probably important to make this distinction on the CMS as well.
- Use primary + secondary when combining a Medium or Severe warning banner + an Info banner.
- Medium or Severe warnings should always be on primary (top) position, and Info should be on secondary position. 
- If combining Medium + Severe warning, the most 'severe' one should be on primary (top) position
- [ ] There is also an option shown for a double severe warning, but if this is too complicated to handle at a CMS level, let's just stack them on top of each other whenever there's a double severe, or any other combinations of 3 (i.e: severe + medium stacked on top, info on secondary position)

src/app/layout.tsx
```diff
@@ -17,12 +18,15 @@ export async function generateMetadata(): Promise<Metadata> {
 export default async function RootLayout({ children }: { children: ReactNode }) {
   const headersList = headers();
   const tokenPrice = await getTokenPrice();
```
BLuEScioN, line 20: Seems like tokenPrice and statusBarContent can be moved further down into the components that directly need the data.

He1DAr: We're doing it here because want to run them on server:
1) token price is caches on server
2) status bar request on client would expose API key

src/app/_components/PageWrapper.tsx
```diff
@@ -56,13 +56,54 @@ function WrapperWithBg({ children }: { children: ReactNode }) {
...
     <>
-      <IncidentsStatusBarWithErrorBoundary />
+      <Flex
+        direction={'column'}
+        width={'100%'}
+        top={'0'}
+        backdropFilter={'blur(10px)'}
+        background={statusBarBg}
+      >
+        <IncidentsStatusBarWithErrorBoundary />
```
BLuEScioN, line 82: Does this not render the status bar?  It's a little confusing to see this then the status bar base being rendered below

src/app/_components/StatusBar/IncidentsStatusBar.tsx
```diff
@@ -1,94 +1,59 @@
...
 
-const incidentImpactSeverity: Record<IncidentImpact, number> = {
-  [IncidentImpact.None]: 0,
-  [IncidentImpact.Minor]: 1,
-  [IncidentImpact.Major]: 2,
-  [IncidentImpact.Critical]: 3,
-};
-
-export function IncidentsStatusBar() {
+export function IncidentsStatusBar(props: FlexProps) {
```
BLuEScioN, line 14: I feel like this file should also contain the IncidentsStatusBarWithErrorBoundary, but rename that to IncidentsStatusBar and name this IncidentsStatusBarBase

BLuEScioN (changes requested):
Some small nits, but overall LGTM

src/app/layout.tsx
BLuEScioN: I don't see server side props anywhere, so curious to know how exactly is this running on the server and not the client?

src/app/_components/PageWrapper.tsx
He1DAr: I agree, will restructure this part.

src/app/_components/StatusBar/IncidentsStatusBar.tsx
He1DAr: not sure I understood this one, can you check if it's still relevant after the latest push?

BLuEScioN: I see that index.tsx loads this dynamically. So I guess it cannot be helped that IncidentsStatusBarWithErrorBoundary is there and not in this file. I still think it's a better convention to suffix components with "withErrorBoundary". Less wordy and it'll be easier to remember the name and look it up. I think it's better to simplify that components name to  IncidentsStatusBar and make that the file name (not sure why it's index.tsx as there are 2 status bars and it only includes one). Then rename the component in this file to IncidentsStatusBarBase.

He1DAr: gotcha, I pushed an update

He1DAr: https://github.com/hirosystems/explorer/compare/1e27d5c40aa74b11c532fbeaa49c81393641bd13..2bb7bd0dea30b1bf97e25de4b76dfd7625596f10

BLuEScioN (approved):

PR #1687, opened 2024-06-20
fix(regressions): fixed a couple bugs

BLuEScioN:
Fixes all bugs mentioned in this Slack thread https://hiropbc.slack.com/archives/C03TU42NL05/p1718708259862839

Biggest fix is for deleting networks.
It wasn't working before because removeCustomNetwork only tried to change customNetworks and not the other list of networks. 
It was confusing having 2 lists of networks, where one was customNetoworks, the other networks, which had networks with property isCustom. To simplify, I removed customNetworks and put everything in networks.
The other tricky thing here is we have to remember both added custom networks and removed custom networks. It's not sufficient to have just a customNetwork cookie. We need two cookies to handle both of these lists. Take for instance a situation where a user removes the old testnet network and adds the naka-1 network. The old testnet needs to initially show up for every user. However, I cannot put it into the added custom networks cookie because I cannot write code the initializes user cookies. Therefore, it will just show by default, unless the user deletes it, then it is added to the removed networks cookie. We then need the added custom networks cookie to track when the naka-1 network is added. These cookies to be kept in sync as well so as not to run into conflicts where a network is simultaneously in the added cookie and the removed cookie

PR #1690, opened 2024-06-21
Fix/mobile menu

### LINKED ISSUE #1556: Mobile UI issue
**Describe the bug**
When on mobile, if you tap the hamburger menu, you can't scroll, so depending on screen size and which submenus are expanded, you might not be able to access the whole menu.

**To Reproduce**
1. Go to https://explorer.hiro.so
2. Tap on the hamburger menu
3. Try to swipe down
4. If the whole menu fits, expand the network or rotate to landscape to see the problem.

**Expected behavior**
I should be able to scroll down to see the rest of the menu.

**Screenshots**
In this screenshot, I can't change the network to testnet because I can't scroll down.

IMG_7225

**Smartphone (please complete the following information):**
 - Device: iPhone 15 Pro
 - OS: iOS 17.3.1
 - Browser: Safari or Chrome

BLuEScioN:
Fixes inability to scroll the mobile menu

Closes #1556

### DIFF
--- a/src/app/_components/NavBar/MobileNav.tsx
+++ b/src/app/_components/NavBar/MobileNav.tsx
@@ -53,6 +53,7 @@ export function MobileNav({
       zIndex={'overlay'}
       padding={6}
       gap={3}
+      overflow="scroll"
     >
       <Flex justifyContent={'space-between'} alignItems={'center'} height={10}>
         <Logo />
--- a/src/app/_components/NavBar/__tests__/__snapshots__/MobileNav.test.tsx.snap
+++ b/src/app/_components/NavBar/__tests__/__snapshots__/MobileNav.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`MobileNav should render correctly 1`] = `
 <DocumentFragment>
   <div
-    class="chakra-stack css-8ccxhr"
+    class="chakra-stack css-tq63gn"
   >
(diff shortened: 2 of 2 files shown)


src/common/utils/utils.ts
```diff
@@ -343,3 +343,18 @@ export function removeTrailingSlash(url?: string) {
   if (!url) return '';
   return url.replace(/\/$/, '');
 }
+
+export function promiseWrapper<T, Args extends any[]>(
```
He1DAr, line 347: Can you explain the purpose of this function?

BLuEScioN: I think this PR was accidentally branched from this [PR](https://github.com/hirosystems/explorer/pull/1687), so it has all those changes. I made a new PR that removed all those changes. The mobile menu PR is a 1 liner. The new PR is [here](https://github.com/hirosystems/explorer/compare/fix/mobile-menu2?expand=1).

But to ansewr the question, this is for my fix for the delete network bug. I saw that you had made addCustomNetwork return a promise, just so you could add a .then hook to it. Instead of muddying the implementation of addCustomNetwork I added this utility function that allows you to take a function, wrap it in a promise that resolves after the function is called, which allows us to add .then hooks to any function we want

He1DAr (approved):

Issue #1550, opened 2024-04-02
Explorer shows error on addresses with no transactions


CharlieC3:
**Describe the bug**
When visiting a valid address, the Explorer will show an error if there are no transactions.
[Example address](https://explorer.hiro.so/address/STGZXM8SYXW1GZ46MRSX6QKQX839C9NPJ0BM5KQ9?chain=testnet).

![Screenshot 2024-04-02 at 4 59 29 PM](https://github.com/hirosystems/explorer/assets/2747302/13ca677a-2816-4332-9553-08b38d8ef678)

This can potentially cause some confusion and make a user think there is a problem with the address when there isn't.
Ideally, the Explorer would be able to determine if it is a valid address and would instead show no transactions instead of an error.

eugeniadigon:
Thanks for flagging, Charlie

I added an empty state on the Address ID page:
https://www.figma.com/file/31q1g27k69z1yXb6r6vrRD/Stacks-Explorer-Working-File?type=design&node-id=7717%3A61388&mode=design&t=WSfTEAv7qSq5xBMC-1

![Address-ID - Empty state](https://github.com/hirosystems/explorer/assets/78865845/b60a83aa-37e8-4f6d-9cc5-5d8e4cf5bf42)

PR #1694, opened 2024-06-25
fix(no-tx-addr): added custom error

BLuEScioN:
Issue https://github.com/hirosystems/explorer/issues/1550

image

We already had a message for when there are no txs. It just wasn't showing. There is a lot of deep logic around displaying the error/no txs message here, from capturing the error at the error boundary to just handling no txs and how that is displayed. I think it would be best to use what we already have for the "no txs message" and don't make any changes until after the redesign.

ST2Y2SFNVZBT8SSZ00XXKH930MCN0RFREB2GQG7CJ


BLuEScioN:
> It works great when navigating directly to the address but not searching. Can we do something about it?

You can still search for it and the search functionality behaves exactly the same as for a address that actually exists. What exactly did you think needs to be changed?

image

image

He1DAr (approved):

src/app/address/[principal]/PageClient.tsx
```diff
@@ -1,17 +1,28 @@
...
 import { StxBalance } from './StxBalance';
 import { TokenBalanceCard } from './TokenBalanceCard';
-import { Wrapper } from './Wrapper';
+
+export function AddressPageLayout(props: GridProps) {
+  return (
+    <Grid
+      gridColumnGap="32px"
+      gridTemplateColumns={['100%', '100%', 'repeat(1, calc(100% - 352px) 320px)']}
+      gridRowGap={['32px', '32px', 'unset']}
```
Comment (He1DAr) [inline src/app/address/[principal]/PageClient.tsx:19]: would you mind using Chakra variables for the gap?

src/common/hooks/useInfiniteQueryResult.ts
```diff
@@ -31,11 +31,14 @@ export function useSuspenseInfiniteQueryResult<T>(
...
-    );
+    const pages = response.data?.pages;
+    const results = pages?.map(page => page.results);
+    const flatResults = results?.flat();
+    const slicedResults = flatResults?.slice(0, limit);
+    return slicedResults || [];
+    // return response.data?.pages
+    //   .map(page => page.results)
+    //   .flat()
+    //   .slice(0, limit) || []
```
He1DAr, line 42:
```suggestion

```

src/app/address/[principal]/PageClient.tsx
Comment (BLuEScioN) [reply src/app/address/[principal]/PageClient.tsx:19]: done

src/app/address/[principal]/TokenBalanceCard/FtBalance.tsx
```diff
@@ -20,13 +20,13 @@ const TokenAssetListItem = dynamic(
...
-  const FTokens = Object.keys(balance.fungible_tokens).filter(
+  const ft = Object.keys(balance.fungible_tokens).filter(
     key => Number(balance.fungible_tokens[key]?.balance) > 0
   );
 
-  return FTokens.length ? (
+  return ft.length > 0 ? (
     <>
-      {FTokens?.map((key, index, arr) => (
+      {ft?.map((key, index, arr) => (
```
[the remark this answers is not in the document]
Comment (BLuEScioN) [reply src/app/address/[principal]/TokenBalanceCard/FtBalance.tsx:29]: done

src/common/hooks/useInfiniteQueryResult.ts
BLuEScioN: done

Issue #1376, opened 2023-12-21
Indicate orphan blocks


andresgalante:
From @obycode 

> Not sure if this is an API or Explorer suggestion (or both), but Jesse and I were just realizing that the explorer page for an orphaned block is a little confusing (e.g. https://explorer.hiro.so/block/0x6c16040b986aba3601624a36d055f93a0ba79070f51ca637642519de38f4c93d?chain=mainnet). There is no indication that it is orphaned, and it shows a Stacks height, but 0 transactions. Should we have some indication about this?

PR #1701, opened 2024-06-27
fix(orphan-blocks): added kv pair to show canonical status

BLuEScioN:
Issue: https://github.com/hirosystems/explorer/issues/1376

I added this canonical key value pair when the block is not canonical to show when the block is orphaned or not. 
@GinaAbrams Do you think this is ok? Or do you want to design this differently?

image

### DIFF
--- a/src/app/block/[hash]/PageClient.tsx
+++ b/src/app/block/[hash]/PageClient.tsx
@@ -1,5 +1,7 @@
 'use client';
 
+import { Modal, ModalContent, ModalOverlay } from '@chakra-ui/react';
+import { Question, X } from '@phosphor-icons/react';
 import dynamic from 'next/dynamic';
 
 import { BtcStxBlockLinks } from '../../../common/components/BtcStxBlockLinks';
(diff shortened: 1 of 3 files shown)


diwakergupta:
> @GinaAbrams Do you think this is ok? Or do you want to design this differently?

Did you mean to tag @eugeniadigon perhaps?

eugeniadigon:
I think it's okay, but we may want to provide some guidance on what this means. For example, do we have an explanation as to why a given block was orphaned?

BLuEScioN:
> I think it's okay, but we may want to provide some guidance on what this means. For example, do we have an explanation as to why a given block was orphaned?

Sometimes two blocks get mined almost simultaneously. Then BTC miners may end up mining blocks for both of those blocks. When one of those blocks gets another block and becomes the longest chain, then miners will switch to supporting that chain, leaving the other mined block orphaned.

eugeniadigon:
> Sometimes two blocks get mined almost simultaneously. Then BTC miners may end up mining blocks for both of those blocks. When one of those blocks gets another block and becomes the longest chain, then miners will switch to supporting that chain, leaving the other mined block orphaned.

Thanks for the explanation. Could we add a (?) icon and a modal that explains this? 
https://www.figma.com/design/31q1g27k69z1yXb6r6vrRD/Stacks-Explorer-Working-File?node-id=8683-8293&t=EXZuHa5wrmnokYXu-1

Block-ID

Block-ID-1

BLuEScioN:
@eugeniadigon 

Does this work?

image

eugeniadigon:
> @eugeniadigon
> 
> Does this work?
> 
> image

Looks good to me 👍

He1DAr (approved):

PR #1718, opened 2024-07-08
feat: advanced search with keywords

He1DAr:
Allow advanced search with keywords `from:ADDRESS | to:ADDRESS | before:YYYY-MM-DD | after:YYYY-MM-DD`

image


src/common/queries/useSearchQuery.ts
```diff
@@ -39,16 +43,80 @@ function nftHistoryToSearchResult(nftHistoryEntry: any, bnsName: string): FoundR
...
+export function parseAdvancedSearchQuery(id: string) {
+  const query: Record<string, string> = {};
+  const isAdvancedSearch = Object.keys(advancedSearchConfig).some(term => id.includes(term));
+  if (!isAdvancedSearch) return query;
+  Object.entries(advancedSearchConfig).forEach(([key, config]) => {
+    const index = id.indexOf(key);
+    if (index !== -1) {
+      const value = id.slice(index + key.length).split(' ')[0];
+      if (!config.isValid(value)) return;
+      query[config.filter] = config.transform(id.slice(index + key.length).split(' ')[0]);
```
BLuEScioN, line 84: Substitute the var value for id.slice(index + key.length).split(' ')[0]

BLuEScioN (approved):

eugeniadigon:
@He1DAr I made some mock-ups to help with the UI of this. You can find them here: https://www.figma.com/design/31q1g27k69z1yXb6r6vrRD/Stacks-Explorer-Working-File?node-id=8765-73039&t=PYb45x43eTeZVGYA-1

Search bar - active
Search results

Let me know what you think

------------

A couple of comments:
- If there are more than 5 results, will we show a button to link to the full results page? There, users can apply more filters. [Here's a mock-up of that page.](https://www.figma.com/design/31q1g27k69z1yXb6r6vrRD/Stacks-Explorer-Working-File?node-id=8662-98088&t=PYb45x43eTeZVGYA-1)
Screenshot 2024-07-09 at 14 40 17

------------

- Ideally, people would type the keyword; once completely typed, it would format itself in the tag style. Example:
Frame-316125192

------------

- Lastly, think we should include a modal to announce this update. @elenananana Would you like to review the copy?
Comms modal

eugeniadigon:
_Bonus:_ This is only a suggestion if we have more time to work on this. The search bar currently is _really_ short in width. It would be nice to expand it to 'full width' upon clicking, so it's easier to type in/view the search terms. It would close/retract when clicking outside of it.

https://www.figma.com/design/31q1g27k69z1yXb6r6vrRD/Stacks-Explorer-Working-File?node-id=8781-51439&t=PYb45x43eTeZVGYA-1

Search results

Example: https://codepen.io/rayblair06/pen/wrdeVy
*Instead of button to bar, it would morph from short bar to long bar

Issue #1078, opened 2023-04-03
Bug: Searching for a fresh address yields nothing: possible fix


314159265359879:
**Describe the bug**
Sometimes searching for an address doesn't yield a result. Visiting the address by editing the normal URL will still reveal the address, usually with no activity. 

Transaction ID 
n.a.

**To Reproduce**
1. Go to explorer.hiro.so
2. Search for a newly created/revealed address
3. See error:
![image](https://user-images.githubusercontent.com/33360391/229617687-21cfaf91-32ba-41ff-a892-ef1d6908c1c7.png)

When instead you visit: https://explorer.stacks.co/address/SP1EYQ4TXYXXP6QY6AM6X15J0KFE4YMYYMRSK23S?chain=mainnet
![image](https://user-images.githubusercontent.com/33360391/229618370-1640c91d-5429-4003-aa82-694e8277c7e6.png)

**Expected behavior**
I am not sure if this is expected behavior, perhaps not the most user friendly? 
Is this happening because the address is not yet considered initialized and that is why it is not yet available in the search index?

Should it be considered initialized as a user searches for it? Can you just default to visiting this kind of page? 
`https://explorer.stacks.co/address/<address>?chain=mainnet`
that would be more user-friendly.

andresgalante:
Thanks for reporting it. I could reproduce it. Here is an example address SP3AGWHGAZKQS4JQ67WQZW5X8HZYZ4ZBWPPJBE6RG

andresgalante:
We think it is an issue with the search API, @rafaelcr can you take a look, please?

https://api.hiro.so/extended/v1/search/SP2AE0JFBMFZFYK52TVE2KR2MFXVJGMSDBVGA8NYT?include_metadata=true&unanchored=true

rafaelcr:
Hmm yeah, unfortunately the API doesn't know this address exists until it has at least 1 transaction so I don't believe there's anything we can do API side here

Could the Explorer maybe detect it is a valid Stacks address and autocomplete that (if the API doesn't return anything)?

andresgalante:
We'll close this issue until the API can inform valid addresses with no tx.

314159265359879:
Is there a related API enhancement issue so we can link to it here?

314159265359879:
@andresgalante I see this is added to a Q3-2023 milestone. I think having a solution for this problem could help resolve issues for users trying to withdraw to a fresh address. Those are sometimes falsely classified as invalid. 

[...]

PR #1719, opened 2024-07-08
fix: added search exception in case an addr is not found

BLuEScioN:
closes [#1078](https://github.com/hirosystems/explorer/issues/1078)


He1DAr:
Could you update the commit message to follow the Conventional Commits specification?

PR #1724, opened 2024-07-09
fix: consistent rendering for 0x

BLuEScioN:


hugocaillard (commented):
Thanks for looking into this @BLuEScioN.
Genuine question about this fix:
Is it worth wrapping the _whole_ page in an extra `div`?
Couldn't the theme be change to add this prop to the `body` or an existing wrapper?

I don't know much about this repo or emotion, so that's really a naive question

BLuEScioN:
> Thanks for looking into this @BLuEScioN. Genuine question about this fix: Is it worth wrapping the _whole_ page in an extra `div`? Couldn't the theme be change to add this prop to the `body` or an existing wrapper?
> 
> I don't know much about this repo or emotion, so that's really a naive question

Good idea. I just combined the flex and styledwrapper

He1DAr (changes requested):
Could you update the commit message to follow the Conventional Commits specification?

He1DAr (approved):

PR #1736, opened 2024-07-19
Feat/holders

BLuEScioN:
Adds holders tab to token page

currently working for ststx - http://localhost:3000/token/SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.ststx-token?chain=mainnet

### DIFF
--- a/src/app/token/[tokenId]/getTokenInfo.ts
+++ b/src/app/token/[tokenId]/getTokenInfo.ts
@@ -1,8 +1,9 @@
 import { FtMetadataResponse } from '@hirosystems/token-metadata-api-client';
 
-import { DEFAULT_MAINNET_SERVER, LUNAR_CRUSH_API_KEY } from '../../../common/constants/env';
+import { LUNAR_CRUSH_API_KEY } from '../../../common/constants/env';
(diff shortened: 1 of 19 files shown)


src/app/token/[tokenId]/getTokenInfo.ts
```diff
@@ -36,10 +36,11 @@ async function getBasicTokenInfo(
   api?: string
 ): Promise<BasicTokenInfo | undefined> {
   const isMainnet = chain === 'mainnet';
-  const isCustomApi = !!api;
+  // const isCustomApi = !!api;
```
Comment (BLuEScioN) [inline src/app/token/[tokenId]/getTokenInfo.ts:39]: This was preventing me from testing locally

src/common/utils/network-utils.ts
```diff
@@ -0,0 +1,7 @@
+export function getApiUrl(chain: string): string | undefined {
```
He1DAr, line 1: we should use api urls from network context or constants as this will introduce another source of truth

BLuEScioN: Have to use constants as this code is being run on the server

He1DAr (approved):

PR #1737, opened 2024-07-19
feat: added a tx notification button to tx page

BLuEScioN:
Added a tx notification button to tx page

### DIFF
--- a/src/app/_components/BlockList/Sockets/use-stacks-api-socket-client.ts
+++ b/src/app/_components/BlockList/Sockets/use-stacks-api-socket-client.ts
@@ -6,11 +6,13 @@ export interface StacksApiSocketClientInfo {
   client: StacksApiSocketClient | null;
   connect: (
     handleOnConnect?: (client: StacksApiSocketClient) => void,
+    handleDisconnect?: (client: StacksApiSocketClient | null) => void,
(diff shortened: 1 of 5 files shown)


BLuEScioN:
In order for this feature to work, notifications have to persist, at the very least, through page refreshes and tab switching. Otherwise, users will switch tabs and the browser will throttle web socket connection, resulting in users not getting notified about their txs. Then, after some time has passed, and the user starts to wonder how their tx isn't complete yet, he will come back to the explorer and refresh the page to check on the status. The status will be complete, but they never got a notification. This will make the user angry and questioning why the notification feature doesn't work.

TLDR, in order to make notifications persist, we need a backend that can manage a persistent websocket connection and updating the client when there is a change in the tx status. The backend will need to handle client tx subscriptions with a Push API service, such that when the tx status changes, the backend can send a messae to the Push API service, which will then send the tx status update message to the client. The client would need to register a service worker that can work in the background in order to pick up this Push message and use the browser's Notification API to inform the user of the new tx status. 

Currently, the explorer does not have a backend, and creating one will be a huge lift and come with significant cost. While notifications would certainly be immensely helpful for users when block times are ~10 minutes, considering that Nakamoto is months away, and fast blocks will render notification updates meaningless, I propose we do not proceed with this feature.

PR #1779, opened 2024-08-27
feat: added sentry

BLuEScioN:
Proof that errors are logging in dev with sourcemaps - https://hirosystems.sentry.io/issues/5801423335/?environment=development&project=4507788896239616&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,14 @@
+# First stage.  This stage is responsible for installing dependencies and building the application.
+# The output of this stage is a built version of your Next.js app, which includes all the static assets and optimized files that will be served to users.
 FROM node:18-alpine AS build
 
-ARG REDIS_URL
-ARG SENTRY_AUTH_TOKEN
-ARG SENTRY_DSN
-ARG SENTRY_LOG_LEVEL=warn
-ARG NODE_ENV=production
-ARG X_API_KEY
-ARG CMS_URL
 ARG RELEASE_TAG_NAME
+ENV NEXT_PUBLIC_RELEASE_TAG_NAME=${RELEASE_TAG_NAME}
 
-# Build args for browser variables
-ARG NEXT_PUBLIC_MAINNET_API_SERVER
-ARG NEXT_PUBLIC_TESTNET_API_SERVER
-ARG NEXT_PUBLIC_LEGACY_EXPLORER_API_SERVER
-ARG NEXT_PUBLIC_DEPLOYMENT_URL
-ARG NEXT_PUBLIC_MAINNET_ENABLED
-ARG NEXT_PUBLIC_DEFAULT_POLLING_INTERVAL
+ARG NEXT_PUBLIC_SENTRY_DSN
+ENV NEXT_PUBLIC_SENTRY_DSN=${NEXT_PUBLIC_SENTRY_DSN}
 
+ARG SENTRY_AUTH_TOKEN
 WORKDIR /app
 
 COPY . .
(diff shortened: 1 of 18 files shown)


sentry.server.config.ts
```diff
@@ -0,0 +1,21 @@
+// This file configures the initialization of Sentry on the server.
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: "https://e1dce4791416146de03ff1642ed719d5@o204651.ingest.us.sentry.io/4507788896239616",
```
He1DAr, line 8: we should move this to env variable so that forks won't use the same dsn

src/app/sentry-example-page/page.tsx
```diff
@@ -0,0 +1,82 @@
+'use client';
```
He1DAr, line 1: is this going to be removed after testing?

He1DAr (changes requested):

src/app/sentry-example-page/page.tsx
BLuEScioN: yes

sentry.server.config.ts
BLuEScioN: done

Dockerfile
```diff
@@ -1,21 +1,15 @@
...
 
-# Build args for browser variables
-ARG NEXT_PUBLIC_MAINNET_API_SERVER
-ARG NEXT_PUBLIC_TESTNET_API_SERVER
-ARG NEXT_PUBLIC_LEGACY_EXPLORER_API_SERVER
-ARG NEXT_PUBLIC_DEPLOYMENT_URL
-ARG NEXT_PUBLIC_MAINNET_ENABLED
-ARG NEXT_PUBLIC_DEFAULT_POLLING_INTERVAL
+ARG SENTRY_AUTH_TOKEN
+ENV SENTRY_AUTH_TOKEN=${SENTRY_AUTH_TOKEN}
```
bgridley, line 12: Is this adding the sentry credentials at build time? I believe they should only be added at run time.

BLuEScioN: Let me check if I can do this

BLuEScioN: That doesn't seem to work @bgridley. https://github.com/hirosystems/explorer/actions/runs/10801775973/job/29962575458

```
 Creating an optimized production build ...
#9 3.180 [@sentry/nextjs - Node.js] Warning: No auth token provided. Will not create release. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
#9 3.180 [@sentry/nextjs - Node.js] Warning: No auth token provided. Will not upload source maps. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
#9 3.239 [@sentry/nextjs - Node.js] Info: Sending telemetry data on issues and performance to Sentry. To disable telemetry, set `options.telemetry` to `false`.
#9 40.61 [@sentry/nextjs - Edge] Warning: No auth token provided. Will not create release. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
#9 40.61 [@sentry/nextjs - Edge] Warning: No auth token provided. Will not upload source maps. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
#9 40.66 [@sentry/nextjs - Edge] Info: Sending telemetry data on issues and performance to Sentry. To disable telemetry, set `options.telemetry` to `false`.
#9 40.78 [@sentry/nextjs - Client] Warning: No auth token provided. Will not create release. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
#9 40.78 [@sentry/nextjs - Client] Warning: No auth token provided. Will not upload source maps. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
#9 40.79 [@sentry/nextjs - Client] Info: Sending telemetry data on issues and performance to Sentry. To disable telemetry, set `options.telemetry` to `false`.
```

BLuEScioN: No, this env var needs to be accessible during the build step, not during runtime

src/app/instrumentation.ts
```diff
@@ -0,0 +1,22 @@
+import * as Sentry from '@sentry/nextjs';
+
+export function register() {
```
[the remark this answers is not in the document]
BLuEScioN: Next.js automatically executes this file on the server side when the application starts up. You don't need to explicitly import or call it anywhere

Dockerfile
BLuEScioN: Just using ARG does work!

CharlieC3: @BLuEScioN That's great! Do you happen to know if the secret is written in plain-text anywhere in the generated code once the app is built in the image? I suppose that would be the last thing to check.

BLuEScioN: I don't think it's written anywhere in plaintext in the image. I pulled the image from docker. Ran it interactively and ran this and nothing came up.
`/app # grep -r "sntrys_eyJpY" /app
`

He1DAr (approved):

PR #1781, opened 2024-08-29
fix: remove metadata calls for holdings

He1DAr:


andresgalante:
What's the status of this PR?

PR #1860, opened 2024-10-15
feat: added new dsign sys colors

BLuEScioN:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adding colors for the new design system. 
I added them as a separate file so that its easy to differentiate the old and new colors and remove the old colors when they are no longer necessary.

https://github.com/hirosystems/explorer/issues/1841

# Checklist:


src/ui/theme/new-colors.ts
```diff
@@ -0,0 +1,83 @@
+export const NEW_COLORS = {
```
andresgalante, line 1: I don't like the prefix "NEW".

What if we create a variable for the prefix instead of being hardcoded and use it everywhere, that way we can dynamically change it or even remove it when we get rid of the old design.

@He1DAr @BLuEScioN  what do you think?

He1DAr: I think it will require a bit of code change whenever we terminate the old theme in both cases. I'm fine with defining a new constant. Maybe adding a TODO comment to describe the situation would be sufficient.

He1DAr (approved):

src/ui/theme/new-colors.ts
BLuEScioN: The NEW is not a prefix being used in the code. That's just the name of the var.

For example:

To use a new color it would look like ` color="var(--stacks-colors-accent-stacks-700)"`

Most of the work of replacing the colors I think would be really easy since the colors have different prefixes. Most of the work would be done by just using find and replace all

andresgalante (approved):

Issue #1843, opened 2024-10-14
Design system: Implement text styles


andresgalante:
https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System-%5BRedesign%5D?node-id=1-11&node-type=canvas&t=yMuH8fZ3dA2N3Fit-0

BLuEScioN:
https://github.com/hirosystems/explorer/pull/1863

PR #1863, opened 2024-10-16
feat: added text styles to theme for redesign

BLuEScioN:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Added the texy styles that we want for the redesign.
Waiting on whether we have the files for the fonts

#1843 

# Checklist:


src/ui/theme/theme.ts
```diff
@@ -121,4 +146,138 @@ export const theme = extendTheme({
     Menu: menuTheme,
     Link: linkTheme,
   },
+  textStyles: {
+    // Regular Variants
+    'text-regular-xs': {
+      fontSize: 'xs',
+      fontWeight: 'regular',
+      letterSpacing: '1%',
+      lineHeight: '16px',
```
andresgalante, line 155: Let's avoid using units for line height. make it just a number which is a ratio of the font size. 

For example if the font size is 18px, and the intended line height is 24, then the line height is `1.33`, no units.

Even though this is less of a problem on modern browsers it still might cause collapsing the lines when users increase the font size.

src/ui/theme/theme.ts
```diff
@@ -121,4 +146,138 @@ export const theme = extendTheme({
     Menu: menuTheme,
     Link: linkTheme,
   },
+  textStyles: {
```
andresgalante, line 149: Are these overrides to the chakra? If so:

1- Why use a number here instead of variables for values? There are for example letterSpacings variables on the system.

2- Do we need all of these overrides? Why not change what's needed and keep the feault value when it's not?

BLuEScioN:
@He1DAr the font faimilies that I have to download don't work in storybook. The ones from google fonts do. I tested that the fonts work on the website, which bundles them correctly. 

If you want me to go back and figure out how to bundle the fonts with storybook I can.

src/ui/theme/theme.ts
BLuEScioN: 1. Done. I am using variable names now
2. I am not overriding the defaults as discussed in our sync. I am just explicitly listing the defaults in our theme so that I can double check values within the code

andresgalante: line-height should also be variables

BLuEScioN: Coming up with names was difficult 😂, but it's done now

andresgalante (approved):

Issue #1786, opened 2024-09-03
Explorer is automatically switching to https when running in local environment


muthusleuths:
**Description**
Explorer is not allowing using http URL for API. When running in local network (for example Umbrel) API is not available over https. This worked fine in the past and now something seems to have changed where by default the connection is attempted over https and then throws an error.

**To Reproduce**
1. Go to Explorer
2. Click on network dropdown
3. Click on "Add Network"
4. Enter "http://umbrel.local:3999" as the API URL
5. See following error "Could not connect to supplied network URL."

**Expected behavior**
The umbrel.local network is added and selected as the API endpoint.

**Screenshots**
<img width="1618" alt="Screen Shot 2024-09-02 at 11 24 26 PM" src="https://github.com/user-attachments/assets/d870e50e-0045-4858-a3d3-15d8e0032462">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Brave

**Additional context**
Running this umbrel project: https://github.com/muthusleuths/stacks-on-umbrel and it pulls the docker image with "latest" tag from https://hub.docker.com/r/hirosystems/explorer/tags which is currently v1.203

muthusleuths:
@BLuEScioN  I am still facing the same error. As you can see from the screenshot my local API runs on my umbrel device and the URL I use to connect is http://umbrel.local:3999 

Does the fix you have in place only work for http://localhost:3999 ?

muthusleuths:
This is from today after running explorer version 1.204.0 on my Umbrel Home
<img width="1622" alt="Screen Shot 2024-09-06 at 12 37 39 PM" src="https://github.com/user-attachments/assets/f23d227e-3642-4d8f-814c-e9463ca31550">

muthusleuths:
@BLuEScioN Can you please take a look at this again. I think the issue is in the code you are looking for the URL to be "localhost" to not force TLS. If you see my screenshots you can see I am connecting to the API in "umbrel.local" endpoint. Not sure if there is an alternate approach we can use here. Perhaps look to pattern match on "local" instead of "localhost" ?

BLuEScioN:
@muthusleuths Can you try again. The fix has been deployed.

muthusleuths:
[...]

PR #1879, opened 2024-10-24
fix: fixed the add network bug

BLuEScioN:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Instead of determining http or https based on whether the urlobj contains localhost, we will base it directly off the protocol that is included in the url passed. If the protocol does not match http or https we will default to https

closes https://github.com/hirosystems/explorer/issues/1786

# Checklist:

image

### DIFF
--- a/src/common/components/__tests__/add-network-form.test.tsx
+++ b/src/common/components/__tests__/add-network-form.test.tsx
@@ -22,6 +22,16 @@ describe('buildCustomNetworkUrl', () => {
       url: 'https://some-custom-url.com/some/path',
       expectedOutput: 'https://some-custom-url.com/some/path',
     },
+    {
+      description: 'builds URL with path',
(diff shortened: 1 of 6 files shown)


src/common/components/__tests__/add-network-form.test.tsx
```diff
@@ -22,6 +22,16 @@ describe('buildCustomNetworkUrl', () => {
       url: 'https://some-custom-url.com/some/path',
       expectedOutput: 'https://some-custom-url.com/some/path',
     },
+    {
+      description: 'builds URL with path',
+      url: 'http://umbrel.local:3999',
```
andresgalante, line 27: Should this be here or is it something you used to test?

andresgalante (approved):
Thanks for adding the extra test

Issue #1624, opened 2024-05-13
Add tenure-height to block height info on the explorer


314159265359879:
**Please review the existing enhancement issues at the link below before creating a new one to ensure you do not create a duplicate request.** 

With the Nakamoto release "tenure-height" is introduced but this is not revealed in the explorer, right? should confirmations have three "block heights" instead:
Stacks block height > tenure-height > bitcoin block height
Tenure-height will be important for users to know when their BNS names expire and any other protocols that rely on "old block-height", I don't think this is revealed anywhere yet.
![image](https://github.com/hirosystems/explorer/assets/33360391/ed566095-ad32-430a-9065-5f0af85805e0)

**Is your feature request related to a problem? Please describe.**
Users will want to know when their BNS names expire they will use "tenure-height"

![image](https://github.com/hirosystems/explorer/assets/33360391/a7878358-b2de-4dd6-80d1-15075986a423)
https://forum.stacks.org/t/bns-nakamoto-upgrade/15579/4

**Describe the solution you'd like**
Have three "block heights" instead:
Stacks block height > tenure-height > bitcoin block height

**Describe alternatives you've considered**
n.a.

**Additional context**
Brice also likes this idea.

andresgalante:
Hi, Since this isn't revealed anywhere yet, Is there any use case for this information at the momemnt?

314159265359879:
Users would need the tenure height to know when their BNS names expire.

andresgalante:
Since it's a feature for BNS, we can show this only on BNS tx-id pages

vicnicius:
+1 here. The tenure block height information can be relevant for many applications, especially those where the interface with the Bitcoin blockchain is relevant. I'm relying on the tenure height because a lot of the functionality in the app I'm working on depends on the `vrf-seed`, and as a result, it is usually more useful to show the user `tenure-heights` instead of stack block heights. If I somehow show this info in my app but this is not present in the explorer it might make for a confusing UX. On the other hand, adding the info doesn't have much of a drawback as far as I can tell.

friedger:
For expiry of BNS or similar burn chain height related logic, the available bitcoin block height is good enough.
However, vrf-seed the tenur height is important.

BLuEScioN:
@zone117x or @obycode Where can this tenure height be found? I do not see it on the blocks endpoints, ie https://api.hiro.so/extended/v2/blocks/{height_or_hash}, etc.

zone117x:
[...]

PR #1885, opened 2024-10-30
feat: added tenure height

BLuEScioN:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adding tenure height to block page.
Our @stacks/blockchain-api-client is outdated. 
```
  api.blocksApi.getBlockByHash({
      hash,
    }),
```
does not have the correct, up-to-date types, so it's not returning tenure height. 
Therefore, I am using fetch for now until we update @stacks/blockchain-api-client and get the new types as well.
I am waiting to do this, as there were some breaking changes and so I think this will be a big task that we should discuss @He1DAr 
I also removed the microblocks page as from my understanding that page is no longer used

https://github.com/hirosystems/explorer/issues/1624#issuecomment-2419947806

# Checklist:


src/app/block/[hash]/PageClient.tsx
```diff
@@ -58,10 +59,13 @@ export default function BlockPage({ params: { hash } }: any) {
...
               <KeyValueHorizontal label={'Mined'} value={<Timestamp ts={block.block_time} />} />
-              <KeyValueHorizontal
-                label={'Transactions'}
-                value={<Value>{block.txs.length}</Value>}
-              />
+              <KeyValueHorizontal label={'Transactions'} value={<Value>{block.tx_count}</Value>} />
+              {(block as any)?.tenure_height !== null && (
+                <KeyValueHorizontal
+                  label={'Tenure Height'}
+                  value={<Value>{(block as any)?.tenure_height}</Value>}
```
Comment (He1DAr) [inline src/app/block/[hash]/PageClient.tsx:66]: Are we able to use the correct type here?

He1DAr (approved):

src/app/block/[hash]/PageClient.tsx
Comment (BLuEScioN) [reply src/app/block/[hash]/PageClient.tsx]: done

PR #2005, opened 2025-02-13
fix: search: return null when entity url exists

He1DAr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Return null when entity url exists to prevent rendering results card after redirect

# Checklist:


BLuEScioN:
@He1DAr Can you please add a PR description and add some context so that it is obvious what this fix is fixing and why the issue occurred in the first place?

BLuEScioN (approved):
Please add a PR description @He1DAr

PR #2024, opened 2025-03-06
feat: redesign - homepage - recent-blocks: BTC view

He1DAr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Recent blocks component - BTC view

# Checklist:
- [x] I have added thorough tests where applicable.

image


src/app/_components/RecentBlocks/BtcBlock.tsx
```diff
@@ -0,0 +1,181 @@
...
+      borderRadius={`calc(var(--stacks-radii-redesign-lg) + ${BORDER_WIDTH}px)`}
+      _hover={{
+        bg: 'var(--stacks-colors-accent-bitcoin-500)',
+        '& .block-height': { textDecoration: 'underline' },
+        '& .stacks-blocks-count': { color: 'textPrimary' },
+      }}
+      cursor={'pointer'}
+      asChild
+    >
+      <a
```
BLuEScioN, line 30: why not a Link component?

src/app/_components/RecentBlocks/BtcBlock.tsx
```diff
@@ -0,0 +1,181 @@
...
+      borderRadius={`calc(var(--stacks-radii-redesign-lg) + ${BORDER_WIDTH + RING_WIDTH * 2}px)`}
+      my={`-${RING_WIDTH * 2}px`}
+      asChild
+      _hover={{
+        bg: 'linear-gradient(to bottom, var(--stacks-colors-alpha-bitcoin-500-alpha-25), var(--stacks-colors-alpha-bitcoin-200-alpha-25))',
+        '& .block-height': { textDecoration: 'underline' },
+      }}
+      role={'group'}
+    >
+      <a
```
BLuEScioN, line 106: Why not <Link/> ?

src/app/_components/RecentBlocks/NewBlockPlaceholder.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { activeNetwork } = useGlobalContext();
+  const lastBlockHeightRef = useRef(newestBtcBlockHeight);
+
+  useEffect(() => {
+    lastBlockHeightRef.current = newestBtcBlockHeight;
+    setHasNewBlocks(false);
+  }, [newestBtcBlockHeight]);
+
+  useEffect(() => {
+    const socketClient = new StacksApiSocketClient({ url: activeNetwork.url });
```
BLuEScioN, line 28: There is a stacks api socket client in global context. Did you consider using that?
  <GlobalContext.Provider
      value={{
        activeNetwork: networks[activeNetworkKey] || {},
        activeNetworkKey, // TODO: rename this to activeNetworkUrl. activeNetwork should be used instead of activeNetworkKey as having the information in both places is redundant
        addCustomNetwork,
        removeCustomNetwork,
        networks,
        apiClient: getApiClient(activeNetworkKey),

        stacksApiSocketClientInfo: {
          client: stacksApiSocketClient,
          connect: connectStacksApiSocket,
          disconnect: disconnectStacksApiSocket,
        },
      }}
    >
      {children}
    </GlobalContext.Provider>
  );

src/app/_components/RecentBlocks/RecentBlocks.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Link } from '../../../ui/Link';
+import { TabsContent, TabsLabel, TabsList, TabsRoot, TabsTrigger } from '../../../ui/Tabs';
+import BitcoinIcon from '../../../ui/icons/BitcoinIcon';
+import StxIcon from '../../../ui/icons/StxIcon';
+import { RecentBtcBlocks } from './RecentBtcBlocks';
+
+export function RecentBlocks() {
+  const network = useGlobalContext().activeNetwork;
+  return (
+    <TabsRoot variant={'primary'} size={'redesignMd'} defaultValue={'btc'} ml={-3}>
```
BLuEScioN, line 15: The ml={-3} seems to be compensating for the pl on the children elements. Why not just remove the pl?

src/app/_components/RecentBlocks/__tests__/formatTimestamp.test.ts
```diff
@@ -0,0 +1,9 @@
+import { formatTimestamp } from '../utils';
+
+describe('formatTimestamp', () => {
```
BLuEScioN, line 3: how does the function perform with other inputs besides a number?

src/app/_components/RecentBlocks/consts.ts
```diff
@@ -0,0 +1,4 @@
+export const BTC_BLOCK_MIN_WIDTH = '184px';
```
BLuEScioN, line 1: put in sizes.ts

src/app/_components/RecentBlocks/utils.ts
```diff
@@ -0,0 +1,16 @@
+export function formatTimestamp(timestamp: number) {
```
BLuEScioN, line 1: I have a very similar function for the tables. I think we should try to consolidate util files. They are starting to get out of hand because there are so many of them

src/ui/TextLink.tsx
```diff
@@ -1,11 +1,11 @@
 'use client';
 
-import { Link, LinkProps, TextProps } from '@chakra-ui/react';
+import { Link, TextProps } from '@chakra-ui/react';
 import { forwardRef } from 'react';
 
 import { Text } from './Text';
 
-export type TextLinkProps = TextProps & LinkProps;
+export type TextLinkProps = TextProps & { href?: string; target?: string };
```
BLuEScioN, line 8: Interesting that LinkProps didn't include these

BLuEScioN (changes requested):
Some small nits. Otherwise, LGTM

src/app/_components/RecentBlocks/RecentBlocks.tsx
He1DAr: it's because we need the parent to have overflow hidden (to trim the list from right) but at the same time we want to show the shadow for the new block placeholder
image

src/app/_components/RecentBlocks/RecentBtcBlocks.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useBurnBlocks } from '../../../common/queries/useBurnBlocksInfinite';
+import { BURN_BLOCKS_QUERY_KEY_EXTENSION } from '../BlockList/consts';
+import { BtcBlock, NewestBtcBlock } from './BtcBlock';
+import { NewBlockPlaceholder } from './NewBlockPlaceholder';
+
+function FadingOverlay() {
+  return (
+    <Box
+      flex={'0 0 auto'}
+      w={['50px', '100px']}
```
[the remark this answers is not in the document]
He1DAr: These ones in particular seem more like case specific numbers rather than core design system constants

src/app/_components/RecentBlocks/__tests__/formatTimestamp.test.ts
He1DAr: The function is typed `formatTimestamp(timestamp: number)` so it shouldn't accept any inputs aside from numbers.

src/app/_components/RecentBlocks/consts.ts
He1DAr: Same for these widths, I think it would help to establish some guidelines around what's considered a design system constant vs a contextual value.

src/app/_components/RecentBlocks/utils.ts
He1DAr: totally agree, we should bring it up in next meeting

src/ui/TextLink.tsx
He1DAr: After update, text and link had conflicting props

BLuEScioN (approved):

src/app/_components/RecentBlocks/NewBlockPlaceholder.tsx
```diff
@@ -0,0 +1,64 @@
...
+export function NewBlockPlaceholder({
+  newestBtcBlockHeight,
+  refetch,
+}: {
+  newestBtcBlockHeight: number;
+  refetch: () => void;
+}) {
+  const [hasNewBlocks, setHasNewBlocks] = useState(false);
+  const lastBlockHeightRef = useRef(newestBtcBlockHeight);
+  useSubscribeBlocks(true, block => {
```
BLuEScioN, line 18: With true hardcoded it seems like this socket connect and block subscription never turns off, but I thought we wanted to turn it off after receiving one block

BLuEScioN (changes requested):
Found one other thing with the block subscription you shoudl look at

Issue #2048, opened 2025-03-14
Bug: Incorrect Cycle Data Display on Signer Dashboard


BowTiedBlox:
**Describe the bug** 
On the **Signer Dashboard**, cycles are completed starting from the latest PoX cycle. However, if a signer has not participated in certain cycles, these cycles are not displayed as **empty/non-participated**. Instead, they are incorrectly populated with data from the previous cycle.  

For example, if **Cycle 103 was not stacked**, it incorrectly displays **Cycle 102’s data** instead of indicating non-participation. This issue propagates, causing **Cycle 102 to display Cycle 101’s data**, and so on.  

**To Reproduce**
1. Visit the following signer’s dashboard: [Signer Explorer](https://explorer.hiro.so/signer/0x02844807121921880119fe05ae47fccb4945a4bb2f840fe7de66e6f32640bc8169?chain=mainnet)
2. The network request for **Cycle 103** returns a **404 error**: [API Request](https://api.hiro.so/signer-metrics/v1/cycles/103/signers/0x02844807121921880119fe05ae47fccb4945a4bb2f840fe7de66e6f32640bc8169)  
3. Instead of marking **Cycle 103 as empty (not stacked)**, it incorrectly fills it with **Cycle 102’s data**.  
4. This error cascades, causing **Cycle 102 to display Cycle 101’s data**.  

**Expected behavior**
- If a signer has **not participated in a cycle**, the dashboard should clearly indicate **non-participation** rather than copying data from the previous cycle.  
- Each cycle should accurately display the correct data without inheriting values from earlier cycles.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d9557224-e832-4a02-b2d8-c2bd5d2a99de)

**Desktop (please complete the following information):**

PR #2057, opened 2025-03-25
fix: fixed the signer cycle mapping bug

BLuEScioN:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The issue with the mapping was it was done after we had removed null values from the array, so the cycle array no longer corresponded to the response array

https://github.com/hirosystems/explorer/issues/2048

# Checklist:

### DIFF
--- a/src/app/signers/data/UseSigner.ts
+++ b/src/app/signers/data/UseSigner.ts
@@ -16,6 +16,8 @@ import {
 
 const SIGNER_QUERY_KEY = 'signer';
 
+type SignerMetricsSignerForCycleWithCycleId = SignerMetricsSignerForCycle & { cycleid: number };
+
 export interface SignerInfo {
   signing_key: string;
   signer_address: string;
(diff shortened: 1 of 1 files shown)


src/app/signers/data/UseSigner.ts
```diff
@@ -84,20 +86,30 @@ export function useSignerStackingHistory(signerKey: string, selectedCycle?: stri
 
   const signerStackingHistoryFiltered = useMemo(() => {
     return signerStackingHistory
+      .map(
```
He1DAr, line 89: Let's do a single pass with better variable naming, maybe something like:
```
const signerStackingHistoryFiltered = useMemo(() => {
  return signerStackingHistory.reduce<SignerMetricsSignerForCycleWithCycleId[]>(
    (acc, signerData, index) => {
      if (!signerData) {
        return acc;
      }

      if ("error" in signerData) {
        return acc;
      }

      acc.push({
        ...signerData,
        cycleid: cyclesToQuery[index],
      });

      return acc;
    },
    []
  );
}, [signerStackingHistory, cyclesToQuery]);
```

BLuEScioN: great idea!

He1DAr (approved):

PR #2071, opened 2025-04-04
feat: Homepage - SSR- Recent blocks

He1DAr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
- Creating new home-redesign page
- Implement SSR for `RecentBlocks` section
- Fix prettier version on CI

#2072

# Checklist:

image

### DIFF
--- a/src/app/home-redesign/context.tsx
+++ b/src/app/home-redesign/context.tsx
@@ -1,0 +1,50 @@
+'use client';
+
+import { ReactNode, createContext, useContext } from 'react';
+
+import { RecentBlocks } from './data';
+
+interface HomePageDataContextType {
+  initialRecentBlocks: RecentBlocks;
+}
+
+const DEFAULT_HOME_PAGE_DATA: HomePageDataContextType = {
+  initialRecentBlocks: {
+    btcBlocks: {
+      results: [],
+      total: 0,
+      limit: 0,
+      offset: 0,
+    },
+    stxBlocks: {
+      results: [],
+      total: 0,
+      limit: 0,
+      offset: 0,
+    },
+  },
+};
+
+const HomePageDataContext = createContext<HomePageDataContextType>(DEFAULT_HOME_PAGE_DATA);
+
+interface HomePageDataProviderProps {
+  children: ReactNode;
+  initialRecentBlocks?: RecentBlocks;
+}
+
+export function HomePageDataProvider({
+  children,
+  initialRecentBlocks = DEFAULT_HOME_PAGE_DATA.initialRecentBlocks,
+}: HomePageDataProviderProps) {
+  const contextValue = {
+    initialRecentBlocks,
+  };
+
+  return (
+    <HomePageDataContext.Provider value={contextValue}>{children}</HomePageDataContext.Provider>
+  );
+}
+
+export function useHomePageData() {
+  return useContext(HomePageDataContext);
+}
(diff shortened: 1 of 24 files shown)


src/app/home-redesign/page.tsx
```diff
@@ -0,0 +1,16 @@
+import { RecentBlocks } from '../_components/RecentBlocks/RecentBlocks';
+import { HomePageDataProvider } from './context';
+import { fetchRecentBlocks } from './data';
+
+export default async function HomeRedesign(props: {
+  searchParams: Promise<Record<string, string>>;
+}) {
+  const searchParams = await props.searchParams;
+  const chain = searchParams?.chain || 'mainnet';
+  const recentBlocks = await fetchRecentBlocks(chain);
```
BLuEScioN, line 10: I don't think this handles the api search param

src/common/components/table/table-examples/BlocksTable.tsx
```diff
@@ -83,7 +83,10 @@ export function BlocksTable() {
     pageSize: 10,
   });
 
-  const response = useBurnBlocks(pagination.pageSize, pagination.pageIndex * pagination.pageSize);
+  const response = useBurnBlocksInfinite(
```
BLuEScioN, line 86: I don't think we need to use infinite here as each request is discreet and no longer relies on getnextpageparam

src/app/home-redesign/context.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { RecentBlocks } from './data';
+
+interface HomePageDataContextType {
+  recentBlocks: RecentBlocks;
+}
+
+const DEFAULT_HOME_PAGE_DATA: HomePageDataContextType = {
+  recentBlocks: {
+    btcBlocks: {
```
BLuEScioN, line 13: I would love to see all of these names also include the prefix 'initial', so 'recentBlocks' becomes 'initialRecentBlocks'. This will make the logic inthe rest of the components easier to follow as we will clearly understand that this is just the initial data being used from the server to make the first render

src/app/home-redesign/data.ts
```diff
@@ -0,0 +1,76 @@
...
+  RECENT_BTC_BLOCKS_COUNT,
+  RECENT_STX_BLOCKS_COUNT,
+} from '../_components/RecentBlocks/consts';
+
+export type UIBtcBlock = Pick<
+  BurnBlock,
+  'burn_block_height' | 'burn_block_time' | 'burn_block_hash' | 'stacks_blocks' | 'total_tx_count'
+>;
+
+export type UIStxBlock = Pick<
```
BLuEScioN, line 16: Just thinking out loud. I think this name is too general. I can see other situations in the explorer where you could want to use the same name for this kind of type. Would be nice to be more specific as this type only pertains to the data needed by the recent blocks UIs.

BLuEScioN (changes requested):
Left a few comments

src/common/components/table/table-examples/BlocksTable.tsx
He1DAr: Make sense, that was just renaming the function though, I suggest we make this refactor in a separate pr.

src/app/home-redesign/data.ts
He1DAr: I've seen `UI` used as a type name prefix to indicate that it's a UI-specific interface. Open to different prefixes.

PR #2084, opened 2025-04-11
feat: addressed tx page feedback

BLuEScioN:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Addressed the tx page feedback provided here 
https://github.com/hirosystems/explorer/pull/2051

# Checklist:


He1DAr (approved):

ginny-d:
Thanks @BLuEScioN 
There are some details that need revision. Let me know if you'd rather hop on a call to go through these:

1. 

- [ ] The filter modal doesn't have the correct padding and spacing
- [ ] It doesn't have the correct overlay color in the background - I _think_ it's just blurred

Screenshot 2025-04-15 at 11 34 50
Screenshot 2025-04-15 at 11 36 27
Screenshot 2025-04-15 at 11 37 00
Screenshot 2025-04-15 at 11 37 20

2. 

- [ ] The filter button has different bottom/top paddings

Screenshot 2025-04-15 at 11 32 09

3. 

- [ ] I don't see this one as completed:

Screenshot 2025-04-15 at 11 35 09

4. 

- [ ] Just checking to confirm this is working as expected: When I apply a filter, I’m seeing only one row per page. As I navigate through the pages, more items appear, but still only one row per page. This is not a good UX - how is this intended to work in the final version?

Screenshot 2025-04-15 at 11 38 39

BLuEScioN:
1. I think you might not be looking at the right deployment. I'm pretty sure I have all the right padding and spacing on the filter modal
image

BLuEScioN:
3. Again, I think you are looking at the wrong deployment
image

BLuEScioN:
4. The transaction type filter in the redesign is operating in the exact same manner as it is in prod today, that is to say it's applying a client side filter, which means it is only being applied to the data that is currently on the page. We agreed that we could keep the functinality the same in order to ship the redesign and address the shortcomings of this implementation in the future. The future implementation would make this filter a server side sort so that if you sorted by tenure change, then you would get a full page of tenure change txs.

ginny-d (approved):

Issue #2123, opened 2025-04-30
[Homepage] Layout is broken on mobile


ginny-d:
Iphone13 Pro, Chrome:

<img width="731" alt="Image" src="https://github.com/user-attachments/assets/98ec9d5a-4ffd-4487-bd69-0ff124b50aca" />

BLuEScioN:
This [PR](https://github.com/hirosystems/explorer/pull/2169) address this

Issue #2122, opened 2025-04-30
[Homepage] Transaction table fixes


ginny-d:
- [x] “View all transactions” link is missing
<img width="707" alt="Image" src="https://github.com/user-attachments/assets/8b6582c2-24cb-406d-ab35-1d190830d124" />

- [x] “To” column is missing

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/e3966920-ca6e-4377-9918-8074adda255b" />

- [x] Since we’re only displaying the last 10 transactions, showing just the time (without the date) is sufficient and more space-efficient. Remove the date from this table

<img width="140" alt="Image" src="https://github.com/user-attachments/assets/50588748-8940-4bf3-ba1f-5d96d141d46b" />

BLuEScioN:
@ginny-d 

Regarding

"Since we’re only displaying the last 10 transactions, showing just the time (without the date) is sufficient and more space-efficient. Remove the date from this table"

What do you think about making this a relative timestamp? It'll be shorter, more concise, and we can add the timestamp on the hover. And I think the relative timestamp approach will be something that'll eventually get reused in the rest of the explorer

BLuEScioN:
PR for this https://github.com/hirosystems/explorer/pull/2169

ginny-d:
> What do you think about making this a relative timestamp? It'll be shorter, more concise, and we can add the timestamp on the hover. And I think the relative timestamp approach will be something that'll eventually get reused in the rest of the explorer

That’s a good idea! I think we should use this on the homepage only, since that’s where we show the most up-to-date and recent activity. In other parts of the Explorer, it might turn into a guessing game (i.e: trying to interpret what "3 days 23h 1m ago" means) and we'd have to rely on a hover state to reveal the exact timestamp, which could add some friction.

On the homepage, could we also use relative timestamps for the Recent Blocks section, please?

BLuEScioN:
Ok will do

PR #2169, opened 2025-05-07
Fix/homepage tx table

BLuEScioN:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Address issues with the tx table on the homepage 

Need to do this
That’s a good idea! I think we should use this on the homepage only, since that’s where we show the most up-to-date and recent activity. In other parts of the Explorer, it might turn into a guessing game (i.e: trying to interpret what "3 days 23h 1m ago" means) and we'd have to rely on a hover state to reveal the exact timestamp, which could add some friction.

On the homepage, could we also use relative timestamps for the Recent Blocks section, please?

closes https://github.com/hirosystems/explorer/issues/2122
closes https://github.com/hirosystems/explorer/issues/2123

# Checklist:


src/ui/ButtonLink.tsx
```diff
@@ -1,26 +1,69 @@
...
+import { ArrowRight } from '@phosphor-icons/react';
 import { ReactNode, forwardRef } from 'react';
 
-import { Button, ButtonProps } from './Button';
 import { Link, LinkProps } from './Link';
 import { NextLink } from './NextLink';
-import { buttonRecipe } from './theme/recipes/ButtonRecipe';
-
-const ButtonLinkBase = forwardRef<
-  HTMLAnchorElement,
```
BLuEScioN, line 12: Perhaps I should keep this and rename the new button link component RedesignButtonLink. Then slowly we'll get rid of the old ButtonLink. Thoughts?

package.json
```diff
@@ -54,6 +54,7 @@
     "bignumber.js": "9.1.2",
     "c32check": "2.0.0",
     "contentful": "11.5.11",
+    "date-fns": "^4.1.0",
```
He1DAr, line 57: pinned version

src/app/_components/RecentBlocks/RecentBlocks.tsx
```diff
@@ -30,8 +43,8 @@ export function RecentBlocks() {
         unmountOnExit
         aria-label="Block view options"
       >
-        <HStack gap={0} pb={3} w={'100%'}>
-          <TabsLabel as="span" id="tab-group-label">
+        <HStack gap={0} pb={4} w={'100%'}>
+          <TabsLabel as="span" id="tab-group-label" whiteSpace={'nowrap !important'}>
```
He1DAr, line 47: why do we need !important here?

src/app/data.ts
```diff
@@ -61,9 +61,9 @@ export interface UIStackingCycle {
 export async function fetchRecentBtcBlocks(chain: string, api?: string) {
```
He1DAr, line 61: changes to this file seem to be out of PR scope

src/ui/theme/recipes/LinkRecipe.ts
```diff
@@ -39,16 +39,19 @@ export const linkRecipe = defineRecipe({
...
+        borderBottom: '2px solid',
+        borderColor: 'redesignBorderSecondary',
         _hover: {
-          textDecoration: 'none',
+          textDecoration: 'none !important',
           borderBottomColor: 'redesignBorderPrimary',
         },
         _disabled: {
-          textDecoration: 'none',
+          textDecoration: 'none !important',
```
He1DAr, line 54: try to avoid !important

BLuEScioN: doesnt seem to be needed anymore

src/app/_components/RecentBlocks/RecentBlocks.tsx
BLuEScioN: doesnt seem to be needed anymore

src/app/data.ts
BLuEScioN: I can put these changes in another PR. The reason I have these changes started with the fact that once I moved the timestamp for txs to relative time, it was obvious that we were not fetching fresh tx data often enough. Initially, I thought it was an issue with the revalidation time. There was an issue there - the revalidation parameter is the cache lifetime of a resource (in seconds). As such, the old revalidation values were much too slow. 
For example, one sample value was 
ONE_MINUTE / 60, where
ONE_MINUTE = 60 * ONE_SECOND, and 
ONE_SECOND = 1000, which
is equal to 60 * 1000 / 60 = 1000 seconds = 16.7 minutes
I think the revalidation values were based on ms and not seconds, so I made the necessary conversions. However, that still did not fix the issue until I changed the cache strategy to "default"

src/ui/ButtonLink.tsx
```diff
@@ -1,26 +1,69 @@
...
+      <Flex alignItems="center" gap={1} className="group">
+        <Text
+          textStyle={buttonLinkSize === 'big' ? 'text-medium-sm' : 'text-medium-xs'}
+          color="textPrimary"
+        >
+          {children}
+        </Text>
+        <Icon
+          w={buttonLinkSize === 'big' ? 3.5 : 3}
+          h={buttonLinkSize === 'big' ? 3.5 : 3}
```
He1DAr, line 33: Can we stick to the size naming used in other components (md, lg, etc) instead of 'big'?
Can we use a recipe to style this component instead of the custom markup?

src/ui/ButtonLink.tsx
He1DAr: I think it's easier to stick with 'redesign' prefix when we start to remove the old code for the sake of consistency, but not a big deal IMO if you're already done with this part.

src/ui/theme/recipes/LinkRecipe.ts
```diff
@@ -39,10 +39,13 @@ export const linkRecipe = defineRecipe({
         },
       },
       buttonLink: {
-        textDecoration: 'none',
+        p: '0 !important',
+        pb: '0.5 !important',
+        textDecoration: 'none !important',
```
He1DAr, line 44: We need to understand why !important is necessary here, it can cause many issue in future.

He1DAr (commented):
Left a few comments.

src/common/utils/time-utils.ts
```diff
@@ -50,3 +51,18 @@ export function formatTimestampTo12HourTime(
...
+  const date = new Date(timestampInSeconds * 1000);
+  const formatted = format(date, formatString);
+  return formatted;
+}
+
+export function formatTimestampToRelativeTime(timestampInSeconds: number): string {
+  const date = new Date(timestampInSeconds * 1000);
+  const relativeTime = formatDistanceToNow(date, { addSuffix: true });
+  return relativeTime;
+}
```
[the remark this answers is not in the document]
BLuEScioN: done

src/ui/theme/recipes/LinkRecipe.ts
BLuEScioN: Important is required because the styles from the default variant are applied and take precedence. IMO this is a bug on Chakra UI's as the defaultVariant should just be the default variant that is applied in the absence of any other variant being specified. If we want to fix this issue ourselves by removing the use of the default variant, I think that should be done in another PR, because it will require changing numerous files

src/common/components/table/Table.tsx
```diff
@@ -309,6 +311,22 @@ export function Table<T>({
...
+                  }
+                  maxW={
+                    header.column.columnDef.maxSize
+                      ? `${header.column.columnDef.maxSize}px`
+                      : undefined
+                  }
+                  w={
+                    header.column.columnDef.size
+                      ? `${header.column.columnDef.size}px`
+                      : 'auto !important'
```
He1DAr, line 327: is !important here necessary?

He1DAr (approved):
just 1 comment regarding using !important

PR #2180, opened 2025-05-09
Search follow-ups

He1DAr:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

# Checklist:
- [ ] I have performed a self-review of my code.


BLuEScioN:
In this video there are 4 things that stand out:
1. There's that weird orange part of the search bar that is non functional and doesn't appear in the designs
2. the signers map shows up on the search bar
3. Hovering over the item gets rid of its background which makes it unclear that you are hovering the item
4. clicking on anything that isn't a link or the searchbar closes the search bar. This is confusing to me as I would expect the overlay to stay open until I close it explicitly either by esc, clicking an X, or clicking outside the searchbar
https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=5274-37555&p=f&m=dev

src/app/_components/Search/Search.tsx
```diff
@@ -497,42 +535,91 @@ function SearchInput({ searchTermFromQueryParams = '' }) {
...
+    } else {
+      document.body.style.overflow = 'auto';
+    }
+  }, [isMobile, isSearchFieldFocused]);
+
+  const Wrapper = useCallback(
+    function Wrapper({ children }: { children: ReactNode }) {
+      if (fullScreen) {
+        return (
+          <Flex
```
BLuEScioN, line 561: Perhaps we should use the same implementation that we are using for the mobile menu, ie
    <Stack
      position="fixed"
      height="full"
      width="full"
      backgroundColor="surfaceTertiary"
      top={0}
      left={0}
      zIndex={'overlay'}
      gap={4}
      padding={6}
      overflow="hidden"
    > I think this would fix the issue where some ui elements are still showing up on top of the searchui

src/app/_components/Search/Search.tsx
```diff
@@ -497,42 +535,91 @@ function SearchInput({ searchTermFromQueryParams = '' }) {
...
+export function Search({ fullScreen = false }: { fullScreen?: boolean }) {
+  const dispatch = useAppDispatch();
   const searchTermFromQueryParams = buildAdvancedSearchQuery(useFilterParams());
   const recentResults = useRecentResultsLocalStorage();
 
   const isSearchFieldFocused = useAppSelector(selectIsSearchFieldFocused);
   const searchTerm = useAppSelector(selectSearchTerm);
   const searchResponse = useSearchQuery(searchTerm, true);
   const isLoading = searchResponse.isLoading;
+  const isMobile = useBreakpointValue({ base: true, lg: false });
```
BLuEScioN, line 547: I think this makes this unable to be ssr. Perhaps we can take the same approach as the mobile menu and make separate the mobile logic into a wrapper around this component. Knowing that it will only be rendered on mobile allows you not to have to use the usebreakpointvalue hook.

see this code for what I am talking about /
const MobileNavBar = ({ tokenPrices, ...props }: { tokenPrices: TokenPrice } & FlexProps) => {
  const { open, onToggle } = useDisclosure();

  return (
    <>
      <SharedMobileNavBar onIconClick={onToggle} icon={<List />} {...props} />
      {open && <MobileNavPage onClose={onToggle} tokenPrices={tokenPrices} />}
    </>
  );
};

src/app/_components/Search/Search.tsx
```diff
@@ -497,42 +535,91 @@ function SearchInput({ searchTermFromQueryParams = '' }) {
...
+                  </Stack>
+                  <SearchResults title={'Recent results'} recentResults={recentResults} />
+                </Stack>
+                {!fullScreen && <QuickLinks />}
+              </>
+            )}
+          </SearchResultsWrapper>
+        )}
+      </HStack>
+      {isSearchFieldFocused && (
```
BLuEScioN, line 607: The implementation might be more straightforward using Chakra UI portals, which we use for modals. Did you look into if that would work?

He1DAr: can you elaborate on the value of adding portal vs just conditionally render the component in this instance?

src/app/_components/Search/Search.tsx
He1DAr: I believe as long a component has a decision to make based on screen size, it has to be client side no matter how we set it up.

BLuEScioN: The issue is that usebreakpointvalue could possible return a value on the server that does not match what it will be on the client, causing a hydration mismatch, multiple renders, and potential UI flashing. When the breakpoint logic is implemented with CSS the server deoesn't have to make a decision and can generate the HTML and CSS and send it to the client where the browser can apply the CSS rules. We also agreed not to use useBreakpointValue, due to these issues. That's why we took it out in the first place.

BLuEScioN (approved):

Issue #2206, opened 2025-05-16
[Homepage] Change relative timestamp to <1 minute ago


ginny-d:
Change relative timestamp to <1 minute ago to remove the horizontal scroll as much as possible

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/dc9edbcc-0588-45c9-9aeb-16a8a71e64cd" />
<img width="199" alt="Image" src="https://github.com/user-attachments/assets/46e95e35-0824-4bf0-b695-fb6c15760c25" />

andresgalante:
We could do the same with `about 1 hour ago` to `~1 hour ago`

PR #2242, opened 2025-05-19
Shorten sub-minute relative times

alexthuth:
**Summary**
This change updates how very recent timestamps are displayed and clarifies prop naming.
The latest transactions page previously rendered “less than a minute ago,” which was verbose and inconsistent with our design guidelines. Issue #2206 requests the more compact string “<1 minute ago.”

**To achieve this:**

- Updated the dayjs locale so that relative times always append “ago” and seconds are rendered as “<1 minute.” 
- Adjusted RelativeTimeDisplay to show “<1 minute ago” whenever the timestamp is under 60 seconds old, while still updating the string every minute for longer intervals.
- Renamed the timestampInMs prop to timestampInSeconds for clarity. This change spans all call sites, including Timestamp and UpdateBar.
- Added a new unit test ensuring toRelativeTime() returns “<1 minute ago” for a timestamp only 30 seconds in the past.
- Documented the workplan for this enhancement under docs/workplans/FEAT-RelativeTimeDisplay.md. The plan outlines the tasks, verification steps, and decision authority for the change.

**Design Decisions**

- Locale configuration: By using a short <1 minute string in the dayjs locale and appending “ago” via the past formatter, we avoid conditional logic in multiple places and maintain consistent phrasing.
- Component prop rename: The previous prop name implied millisecond input but all timestamps were provided in seconds. Renaming it to timestampInSeconds clarifies the intent and prevents misuse.
- Test coverage: A targeted unit test confirms that the helper toRelativeTime produces the new string for very recent times, safeguarding against regressions.

**Files Changed**

- src/common/utils/utils.ts
- src/common/components/RelativeTimeDisplay.tsx
- src/common/components/Timestamp.tsx
- src/app/_components/BlockList/UpdateBar.tsx
- src/common/utils/__tests__/relative-time.test.ts
- docs/workplans/FEAT-RelativeTimeDisplay.md

Testing

- ✅ pnpm lint
- ✅ pnpm test:unit
- ❌ pnpm build (fails when downloading Google Fonts on node:18-alpine because the container lacks CA certificates)

To resolve the build failure, install ca-certificates in the Dockerfile’s build stage before running pnpm build.

This PR addresses issue #2206 by introducing concise relative time labels and improving prop clarity.

### DIFF
--- a/docs/workplans/FEAT-RelativeTimeDisplay.md
+++ b/docs/workplans/FEAT-RelativeTimeDisplay.md
@@ -1,0 +1,23 @@
+Task ID: FEAT-RelativeTimeDisplay
+Problem Statement: Existing relative time formatting displays long strings below a minute and the prop name timestampInMs is misleading. Need to display '<1 minute ago' and rename prop for clarity.
+Components Involved: RelativeTimeDisplay component, utils.ts locale config, UpdateBar, Timestamp, unit tests.
+Dependencies: dayjs locale configuration, React components.
+Implementation Checklist:
+- [ ] Update dayjs locale relativeTime.s to '<1 minute' and past to always append ' ago'.
+- [ ] Modify RelativeTimeDisplay to render '<1 minute ago' when diff < 60.
+- [ ] Rename prop timestampInMs to timestampInSeconds across the codebase.
+- [ ] Update affected components and tests.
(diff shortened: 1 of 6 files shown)


seconds-0:
SS from the build:

image

Issue #2222, opened 2025-05-19
[Signers] Remove Stacking section


ginny-d:
Since this section is not showing accurate data and we already have a similar—but visually different—component on the homepage, let's remove it from the Signers page to avoid confusion.

<img width="669" alt="Image" src="https://github.com/user-attachments/assets/65d1a047-aa2b-4023-a28f-d14979d56e28" />

PR #2245, opened 2025-05-19
Remove stacking section from signers page

alexthuth:
**Summary**

- Removed the stacking section from the Signers page so it now focuses solely on signer data. The simplified header is shown in SignersHeader.tsx where only signer distribution and map elements remain
- Updated the skeleton file to match the new layout, ensuring the placeholder header renders only signer distribution data
- Dropped the token price dependency and simplified the client component and route; the page now simply renders <SignersHeader /> without fetching additional data

**Implementation Notes**

- Stacking-related card components were removed from the header layout but kept in the repository for potential reuse.
- Signers page loading states were adjusted to match the trimmed component tree.

The change addresses issue #2222.

**Files Changed**

- src/app/signers/PageClient.tsx
- src/app/signers/SignersHeader.tsx
- src/app/signers/page.tsx
- src/app/signers/skeleton.tsx

**Testing**

-  ✅ pnpm lint (with warnings)
-  ✅ pnpm test:unit
- ❌ pnpm build (failed while downloading fonts in CI due to missing CA certificates)

### DIFF
--- a/src/app/signers/PageClient.tsx
+++ b/src/app/signers/PageClient.tsx
@@ -2,18 +2,17 @@
 
 import { Flex } from '@chakra-ui/react';
 
-import { TokenPrice } from '../../common/types/tokenPrice';
 import { PageTitle } from '../_components/PageTitle';
 import { SignersHeader } from './SignersHeader';
 import SignersTable from './SignersTable';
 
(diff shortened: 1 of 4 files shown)


alexthuth:
image
image


Appears to have satisfied the ticket

BLuEScioN (approved):

PR #2247, opened 2025-05-19
fix: show contract deploy filter label

alexthuth:
## Summary
- map active transaction type labels using `getTxTypeLabel`
- add regression test for contract deploy filter label
- document workplan

## Testing
- `pnpm lint` *(fails: ESLint couldn't find config)*
- `pnpm test:unit` *(fails: jest not found)*
- `pnpm build` *(fails: chakra command not found)*


alexthuth:
image

New change showing "Contract Deploy" matching title

Issue #2221, opened 2025-05-19
[Transactions] Remove the N/A tag when there is no price value data available


ginny-d:
Remove the N/A tag when there is no price value data available, instead of displaying N/A.

<img width="664" alt="Image" src="https://github.com/user-attachments/assets/6547c148-0bef-46f3-95c5-f3b085481cd3" />

andresgalante:
yes! This already broke before, let's fix it again and add a safeguards it doesn't show if it's N/A

PR #2249, opened 2025-05-19
Fix hide N/A price badge

alexthuth:
## Summary
- remove N/A fallback from `getUsdValue`
- hide `StxPrice` and `StxPriceButton` when price data is missing
- add unit test verifying the price component doesn't render without price data

## Testing
- `pnpm lint` *(fails: ESLint couldn't find configuration)*
- `pnpm test:unit` *(fails: jest not found)*
- `pnpm build` *(fails: command "chakra" not found)*

Addresses https://github.com/hirosystems/explorer/issues/2221

### DIFF
--- a/src/__tests__/components/StxPrice.test.tsx
+++ b/src/__tests__/components/StxPrice.test.tsx
@@ -1,0 +1,22 @@
+import { render } from '@testing-library/react';
+import { StxPrice } from '../../common/components/StxPrice';
+import { useStxPriceForTx } from '../../common/components/StxPriceButton/useStxPriceForTx';
+import { useGlobalContext } from '../../common/context/useGlobalContext';
+
+jest.mock('../../common/components/StxPriceButton/useStxPriceForTx');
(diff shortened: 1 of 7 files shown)


alexthuth:
image

No longer showing N/A

ginny-d:
I still see it on transaction ID pages: https://explorer.dev.hiro.so/txid/0x395518806edb78e362d74af00de5596dc11bb01b7589c3b489eb1297b81148c2?chain=mainnet 

Screenshot 2025-05-20 at 11 08 33

BLuEScioN:
There is something wrong with the price endpoint we are using, which is causing a lot of these N/As to show up.

@ginny-d Just raising attention to the fact that is change may make some UX a little awkward, ie you change the value basis filter and the prices just disappear because the there is no price information.

@alexthuth This change would have had the unintended side effect of removing the N/A from the balance widget on address pages
https://files.slack.com/files-pri/T02BM1EP6-F08T0HKD1MZ/image.png

Issue #2216, opened 2025-05-19
[Homepage] Timestamps in the Recent Blocks widget appear inconsistent between mainnet and testnet.


ginny-d:
Compare mainnet vs. testnet timestamps to see the discrepancy. Likely related to how data is fetched or cached during the network switch.

Also when switching to testnet, different timestamps briefly flash before updating — suggesting a rendering or state issue.

https://github.com/user-attachments/assets/70844646-d856-4a4c-8530-1a587f9a2b8b

PR #2250, opened 2025-05-19
Block Timestamps are different between testnet and mainnet

alexthuth:
## Problems
- Timestamps in recent blocks widget are inconsistent between mainnet and testnet
- When switching between mainnet and testnet, timstamps flash before updating

## Summary
- Cleared the React Query cache whenever the active network changes, preventing stale block data from appearing after a network switch
- Introduced a helper to build query keys with the active network identifier, ensuring per-network caching for burn block requests
- Added unit tests verifying that the generated query keys include the active network identifier to guard against regressions

## Testing
- `pnpm lint`
- `pnpm test:unit`
- `pnpm build` *(fails: Failed to fetch `Instrument Sans` from Google Fonts)*

Fixes: https://github.com/hirosystems/explorer/issues/2216

### DIFF
--- a/docs/workplans/2216-Timestamps.md
+++ b/docs/workplans/2216-Timestamps.md
@@ -1,0 +1,38 @@
+# 2216 - Homepage Block Timestamps
+
+## Problem Statement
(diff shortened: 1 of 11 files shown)


alexthuth:
Closing this PR because I am opening a new one that does not have chakra build errors

PR #2251, opened 2025-05-20
Fix network cache key for block txs

alexthuth:
## Summary
- cache block tx results per-network
- test query key helper
- add network switch e2e spec
- ensure signer map components run on client
- add workplan for network switch query key

## Testing
- `pnpm lint`
- `pnpm test:unit`
- `NEXT_FONT_IGNORE_DOWNLOAD_ERRORS=1 pnpm build`
- `pnpm test:e2e` *(fails: Command failed with exit code 1)*

### DIFF
--- a/docs/workplans/001-NetworkSwitchQueryKey.md
+++ b/docs/workplans/001-NetworkSwitchQueryKey.md
@@ -1,0 +1,43 @@
+Task ID: 001
+
+Problem Statement:
+The Recent Blocks timestamps briefly show stale data after switching networks because the useBlockTxsInfinite query cache is keyed only by blockHash. React Query reuses cached data for a different network until refetched, leading to visual glitches. Additionally, several components under src/app/signers require the 'use client' directive but currently omit it, causing runtime errors when Chakra hooks run on the client.
+
+Components Involved:
+- src/common/queries/useBlockTxsInfinite.ts
+- src/app/signers/SignersMap.tsx
+- src/app/signers/SignersMapComponent.tsx
+- src/app/signers/CurrentCycleCard.tsx
+- src/common/queries/__tests__
+- e2e tests
+
+Dependencies:
+- React Query
+- Global context for network state
+
+Implementation Checklist:
+- [x] Add helper getBlockTxsQueryKey(blockHash, activeNetworkKey).
+- [x] Use useGlobalContext within hooks to include network key in queryKey.
+- [x] Add unit test validating helper output.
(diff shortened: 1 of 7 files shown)


docs/workplans/001-NetworkSwitchQueryKey.md
```diff
@@ -0,0 +1,43 @@
+Task ID: 001
```
BLuEScioN, line 1: remove

e2e/network-switch.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { expect, test } from '@playwright/test';
```
BLuEScioN, line 1: This test wasn't run by Codex and it doesn't pass. I believe we don't have any e2e tests running right now

src/common/queries/__tests__/useBlockTxsInfinite.test.ts
```diff
@@ -0,0 +1,7 @@
+import { getBlockTxsQueryKey } from '../useBlockTxsInfinite';
```
BLuEScioN, line 1: Don't think this function needs a test. It's straightforward

e2e/network-switch.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  await page.waitForSelector('[aria-label^="Block timestamp"]');
+  return page.locator('[aria-label^="Block timestamp"]').first().innerText();
+}
+
+test.describe('network switch', () => {
+  test('block timestamps update on network change', async ({ page }) => {
+    await page.goto('/');
+    const mainnetTimestamp = await getFirstBlockTimestamp(page);
+
+    await page.click('#settings-popover button');
```
BLuEScioN, line 14: This doesn't exist. The AI hallucinated

e2e/network-switch.spec.ts
BLuEScioN, line 1: In general, I don't really like creating an E2E test for this. It's pretty complex fo what it's testing and it's not obvious from the implementation what the E2E test is testing. The way I see it, the fix about making sure that react query caches the query for block txs correctly by incorporating the network key in the query key. We don't want to test react query and whether it correctly caches the query based on the query key provided, so the only thing to test is that the query key is included in the query provided to the use query hook, and we know it is because it is a required argument to the function for generating the query key, so I don't think a test is really necessary here

BLuEScioN (changes requested):
Deletes workplan and tests. The E2E test is no good.

PR #2335, opened 2025-06-11
ci: add gpg to stg and prd deploys

deantchi:
Adds signed commits to `stg` and `prd` deployments.


He1DAr:
@deantchi do we still need this PR?

PR #2338, opened 2025-06-11
feat: tx fee estimates

### LINKED ISSUE #2309: Create a mechanism to estimate fees
1. Showing a true fee estimate like wallets do isn't feasible, since wallets show estimates based on real txs users are preparing to broadcast. The Explorer doesn't have access to that context, and replicating that logic would be complex. I'm also unsure of the value of showing that in the Explorer environment, given users will likely check fees directly in their wallets anyway.
2. Displaying "fees currently in the mempool" offers little value, as those transactions have already happened and it's not a very meaningful piece of information.
3. Users may still visit the Explorer looking for fee information, and showing nothing at all could feel like a gap or create confusion

He1DAr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Update tx fee estimates to calculate fee using sample txs.

closes #2309

# Checklist:


src/app/_components/FeeSection/index.tsx
```diff
@@ -121,14 +124,25 @@ function getFeeDescription(txType: keyof MempoolFeePriorities) {
...
+  const { feeEstimates } = useHomePageData();
+
+  const feeEstimatesMap = {
+    all: feeEstimates.averageFees,
+    smart_contract: feeEstimates.contractDeployFees,
+    token_transfer: feeEstimates.tokenTransferFees,
+    contract_call: feeEstimates.contractCallFees,
+  };
+
+  const txTypeFees = ['all', 'smart_contract', 'token_transfer', 'contract_call'] as Array<
```
BLuEScioN: Can place this outside the function

src/app/page.tsx
```diff
@@ -23,22 +24,37 @@ export default async function HomeRedesign(props: {
...
+    initialTxTableData,
+    mempoolStats,
+    sampleTxsFeeEstimate,
+  ] = await Promise.all([
+    getCurrentStxPrice(),
+    fetchRecentBlocks(chain, api),
+    fetchCurrentStackingCycle(chain, api),
+    fetchRecentUITxs(chain, api),
+    fetchUIMempoolStats(chain, api),
+    getSampleTxsFeeEstimate(chain as 'mainnet' | 'testnet', api),
```
BLuEScioN, line 40: is there a better type we can use here?

BLuEScioN (approved):
Great work! This looks really nice. Approving with some small nits

src/app/page.tsx
He1DAr: The thing is leather library expects a different chain type ('mainnet' | 'testnet' | 'devnet'...), so I had to cast our chain type to match.

Issue #2218, opened 2025-05-19
[Filters] Add shortcut to remove individual filters


ginny-d:
Currently, users can only clear all filters at once using the “Clear All” option. It would improve UX to allow removing individual filters directly.

Proposal: Add an “×” icon on hover for each active filter pill/chip, enabling users to remove filters one by one without resetting everything.

BLuEScioN:
Can you add the design for this please? @ginny-d

ginny-d:
@BLuEScioN I updated the designs here: https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=2340-12810&t=9rZ4FnzZdv7ewzTO-1

I would also recommend changing "Clear filters" for "Clear all filters" just to make it more explicit, now that we have both functionalities

BLuEScioN:
@ginny-d What is the design for this on mobile?

ginny-d:
Oops, sorry I omitted that

Let's follow the same pattern as desktop if possible. And can we add a "Clear all filters" button as well?
<img width="477" alt="Image" src="https://github.com/user-attachments/assets/b26d53a9-0842-42e2-8bff-c2d9973879ef" />



https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=3160-11275&t=MNs6Mcm3KMgc10KP-1

PR #2351, opened 2025-06-14
Feat/remove individual filters

BLuEScioN:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds UI onto the filters to remove them individually

closes  https://github.com/hirosystems/explorer/issues/2218

# Checklist:


ginny-d:
Thanks @BLuEScioN, it looks & works great!

Just one small request if possible:
Changing the X icon color from `icon-tertiary` to `icon-secondary`.

Reasoning:
While Figma specifies icon-tertiary, the implementation uses the regular icon weight instead of bold (used in Figma). This results in reduced contrast and visibility. Updating to `icon-secondary` should help align the visual clarity with the original design intent.

src/components/ui/popover.tsx
```diff
@@ -49,6 +49,13 @@ export const PopoverCloseTrigger = React.forwardRef<
   );
 });
 
+export const PopoverTriggerWithClear = React.forwardRef<
+  HTMLButtonElement,
+  ChakraPopover.TriggerProps
+>(function PopoverTriggerWithClear(props, ref) {
+  return <ChakraPopover.Trigger {...props} ref={ref} />;
+});
+
```
He1DAr: would it work with something like:
```suggestion
export const PopoverTrigger = ChakraPopover.Trigger;
```

src/common/components/table/filters/table-filters-utils.tsx
```diff
@@ -60,6 +60,25 @@ export const useDateFilterHandler = (useShallow: boolean = true) => {
...
+  const params = new URLSearchParams(searchParams);
+  const searchParamsWithoutDateFilter = removeDateFilterFromSearchParams(params);
+  if (useShallow) {
+    return () => shallowRouter.replace(null, '', `?${searchParamsWithoutDateFilter.toString()}`);
+  } else {
+    return () => router.replace(`?${searchParamsWithoutDateFilter.toString()}`);
+  }
+};
+
 export const getAddressFilterParams = (
```
He1DAr: I'm a bit unclear on the intent of this function, name suggests it's a getter but it can set or delete a params.

src/common/components/table/filters/table-filters-utils.tsx
```diff
@@ -60,6 +60,25 @@ export const useDateFilterHandler = (useShallow: boolean = true) => {
   };
```
He1DAr: This file seems to have a repeated pattern across all filter handlers, each one follows similar structure: building URLSearchParams, applying changes, and updating the URL (router.replace or shallowRouter.replace)

Would you consider abstracting this into a generic testable utility (e.g. useQueryUpdater) that takes a param mutation function and handles the routing logic? It might help reduce duplication and make it easier to add new filters in the future. Also maybe putting all filter keys/query param into enum for better typing.

src/common/components/table/filters/transaction-type-filter/TransactionTypeFilterAccordionItem.tsx
```diff
@@ -1,48 +1,75 @@
...
+  //       p={1.5} // I think there is a bug on Chakra that's causing the padding here to be applied to 2 divs surrounding the content
+  //     >
+  //       <TransactionTypeFilterForm
+  //         open={open}
+  //         onSubmit={onSubmit}
+  //         defaultTransactionType={transactionType}
+  //       />
+  //     </AccordionItemContent>
+  //   </AccordionItem>
+  // );
```
He1DAr:
```suggestion
```

src/common/components/table/filters/transaction-type-filter/TransactionTypeFilterPopover.tsx
```diff
@@ -1,30 +1,37 @@
...
         placement: 'bottom-start',
         offset: { mainAxis: TAB_HEIGHT_ADJUSTMENT },
         sameWidth: true,
       }}
       trigger={(open, setOpen) => (
-        <TransactionTypeFilterTriggerText open={open} transactionType={defaultTransactionType} />
+        <TransactionTypeFilterTrigger
+          open={open}
+          transactionType={defaultTransactionType}
+          clearFilterHandler={clearFilterHandler}
```
He1DAr: Can `useTxTableFilters` be used in component directly to avoid prop drilling?

src/common/components/table/filters/table-filters-utils.tsx
```diff
@@ -94,6 +119,21 @@ export function useAddressFilterHandler(useShallow: boolean = true) {
...
 }
 
+export const useClearAddressFilterHandler = (useShallow: boolean = true) => {
```
He1DAr: It looks like all clear hooks follow same pattern, we could maybe have 1 hook:
```
export const useClearFilterHandler = (removeFn, useShallow) => {
  const searchParams = useSearchParams();
  const shallowRouter = useShallowRouter();
  const router = useRouter();

  return () => {
    const params = new URLSearchParams(searchParams);
    const updatedParams = removeFn(params);
    const query = `?${updatedParams.toString()}`;
    useShallow
      ? shallowRouter.replace(null, '', query)
      : router.replace(query);
  };
};
```
then we reuse it like:
```
export const useClearDateFilterHandler = (useShallow) =>
  useClearFilterHandler(removeDateFilterFromSearchParams, useShallow);

export const useClearAddressFilterHandler = (useShallow) =>
  useClearFilterHandler(removeAddressFilterFromSearchParams, useShallow);
  
....
```

src/common/components/table/filters/table-filters-utils.tsx
He1DAr: I think adding a unit test suite for this file would be really valuable to help catch edge cases and give us confidence updating it in future.

He1DAr (changes requested):
Great work! Left a few comments around potential abstractions and testing.

src/common/components/table/filters/transaction-type-filter/TransactionTypeFilterAccordionItem.tsx
BLuEScioN: done

src/common/components/table/filters/transaction-type-filter/TransactionTypeFilterPopover.tsx
BLuEScioN: The reason I am not doing that is because I want these filter components to be more abstracted so that they are reusable across the explorer, on other pages, such as the transaction id page, or the blocks page. If I use the useTxTableFiters context hook in them then I am tying their implementation to the TxTable.

src/components/ui/popover.tsx
BLuEScioN: This isn't used anywhere. I think I was just playing around with an idea

src/common/components/table/filters/table-filters-utils.tsx
BLuEScioN: I see your point. Those functions have been removed, and I have abstracted the query manipulation functionality into a hook called useQueryUpdater, and I have simplified the logic for handling query params for each filter into individual filter mutator functions.

src/common/components/table/filters/table-filters-utils.tsx
BLuEScioN: done

BLuEScioN: done

PR #2460, opened 2025-08-11
feat: redesign: blocks page

He1DAr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
- redesign: blocks page
- migrate to v2 blocks endpoints

# Checklist:

image
image


src/ui/Badge.tsx
```diff
@@ -127,20 +127,20 @@ export const DefaultBadgeIcon = ({
       h={4.5}
       w={4.5}
     >
-      <Icon h={3} w={3} color={color}>
```
BLuEScioN, line 130: Did you clear this with Ginny?

BLuEScioN (approved):
Left a question about the change to the badge styles

ginny-d:
@He1DAr thanks for sharing this, here are some last revisions:


- It looks like the transaction type tag has been wrongly affected by a new style. Left is the Vercel preview and right is the current Explorer (expected font style)
Screenshot 2025-08-18 at 21 41 14
Screenshot 2025-08-18 at 21 41 56

--------
- The underline color of the Block tag doesn't really need to change between default and hover state, it should always be black like here:
Screenshot 2025-08-18 at 21 43 01

---------- 
- Is it possible to add this hover state that triggers when hovering on the Bitcoin row (Stacks blocks view)?
Screenshot 2025-08-18 at 21 44 02

----------
- Minor detail on mobile: the top margin of the "horizontal scroll" bar is different than the bottom marging. Expected should be both like the bottom margin
- Also, is it possible to not wrap in two lines the hash?

photo_2025-08-18_21-46-53
photo_2025-08-18_21-46-55 (2)

src/ui/Badge.tsx
He1DAr: I updated this to only affect blocks page

PR #2474, opened 2025-08-18
Feat/call function flow

BLuEScioN:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds redesign UI for calling a function 

# Checklist:


src/app/txid/[txId]/PageClient.tsx
```diff
@@ -70,8 +70,9 @@ function TransactionIdPage() {
 
   if (tx?.tx_type === 'coinbase') return <CoinbasePageRedesign tx={tx} />;
 
-  if (tx?.tx_type === 'contract_call' && isRedesign) return <ContractCallPageRedesign tx={tx} />;
-  if (tx?.tx_type === 'contract_call') return <ContractCallPage tx={tx} />;
+  if (tx?.tx_type === 'contract_call') return <ContractCallPageRedesign tx={tx} />;
+  // if (tx?.tx_type === 'contract_call' && isRedesign) return <ContractCallPageRedesign tx={tx} />;
```
Comment (BLuEScioN) [inline src/app/txid/[txId]/PageClient.tsx]: revert this change

src/common/utils/sandbox.ts
```diff
@@ -12,8 +18,24 @@
 export function showFn(contractId: string, abiFn: any) {
   return (
     abiFn.access !== 'private' &&
-    (abiFn.name !== 'stack-increase' ||
```
BLuEScioN: It didn't make sense that we were making it invalid for function names to be "stack-increase". I adjusted the check to allow for the function name to be this, so long as the contract is not of the contracts listed

ginny-d:
Hey @BLuEScioN thanks for sharing this!

Here's my revision:

-----------------------
1.
Screenshot 2025-08-25 at 15 19 44
-----------------------

-----------------------
-----------------------
-----------------------
-----------------------
-----------------------
-----------------------
-----------------------
-----------------------
-----------------------
-----------------------

ginny-d:
Hey @BLuEScioN thanks for sharing this!

Here's my revision:

-----------------------

- [x] 1. When a (#) is present, tab styles differ between Token Transfer and Contract Call pages.
✅ Expected: Match styling from Token Transfer

Screenshot 2025-08-25 at 15 19 44

-----------------------
- [x] 2. Block height tags: Current tags don’t match Figma, and hover styling differs from the Blocks page.
✅ Expected: Use Primary tag style.
[Figma reference](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=4416-9439&t=NzxSUqg9taECjv4Q-1)

Screenshot 2025-08-25 at 15 25 08


-----------------------
- [x] 3. Underline styling is inconsistent across rows.
✅ Expected: Match the styling of the first row.

Screenshot 2025-08-25 at 15 26 02

-----------------------
- [x] 4.Contract tag: Different style from contract tags already in use in transaction tables.
✅ Expected: Reuse the same contract tag style as the transactions table for consistency.

Screenshot 2025-08-25 at 15 26 49

-----------------------
- [x] 5. Tag label
Current: "function call"
✅ Expected: "function called"

Screenshot 2025-08-25 at 15 03 19

-----------------------
- [x] 6. Success/confirmed + deny/mode badges: Badge styles differ across implementations.
✅ Expected: Match Figma component.
[Figma reference](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=4416-9426&t=NzxSUqg9taECjv4Q-1)

Screenshot 2025-08-25 at 15 28 54


-----------------------

- [x] 7. Function tags
Currently styled in Instrument Sans, inconsistent with other tags.
✅ Expected: Use the Simple Tag component (Matter Mono).
[Figma reference](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=4427-2462&t=NzxSUqg9taECjv4Q-1)

Screenshot 2025-08-25 at 15 31 31
Screenshot 2025-08-25 at 15 35 49

-----------------------
- [x] 8. Let's use a simple empty state, as this is the one used for search results. Let's use "No post-conditions to show"
Screenshot 2025-08-25 at 15 06 08


-----------------------
- [x] 9. Inconsistent events table styling:
On Token Transfer: table border missing.
On Contract Call: filters are placed inside the border.
✅ Expected: Align styling across pages like shown in Figma.

Screenshot 2025-08-25 at 15 34 38

-----------------------
- [x] 10. Copy button hover state has inconsistent styling.
✅ Expected: Match the copy button next to the function tag.

[Figma reference](https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=5448-20546&t=VKAuKNP3xlW9xn3B-1)
Screenshot 2025-08-25 at 15 38 11

-----------------------
- [x] 11. Function called table column are misaligned.
✅ Expected: Align column starts consistently.

Screenshot 2025-08-25 at 15 38 50


-----------------------
- [x] 12. Code editor height feels restrictive. Can we expand vertically to make better use of viewport space?
Screenshot 2025-08-25 at 15 09 08

-----------------------
- [x] 13. Scroll bug in expanded code editor: When placing the pointer on the code editor, it's not possible to scroll.
Screenshot 2025-08-25 at 15 09 32

-----------------------
- [x] 14. Function called table headers & index tags:
Headers don’t align with row starts when the table is long.
Index tag is set in Instrument Sans instead of Matter Mono (like the Simple Tag component).
✅ Expected: Align headers to the top of the row and update typography.
Screenshot 2025-08-25 at 15 42 42

-----------------------

Thanks, and let me know if you have any questions!

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+          return tupleAcc;
+        },
+        {} as Record<string, string | number>
+      )
+    : '';
+  return (
+    <FieldArray name={name}>
+      {({ push, remove }) => (
+        <Stack gap={4} w="full">
+          <Text textStyle="text-medium-sm" as={'label'}>
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: could maybe use Field.Label like the other components

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ? listTuple.reduce(
+        (tupleAcc, tupleEntry) => {
+          tupleAcc[tupleEntry.name] = '';
+          return tupleAcc;
+        },
+        {} as Record<string, string | number>
+      )
+    : '';
+  return (
+    <FieldArray name={name}>
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: we should wrap with Field.Root

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+      {({ push, remove }) => (
+        <Stack gap={4} w="full">
+          <Text textStyle="text-medium-sm" as={'label'}>
+            {name}
+          </Text>
+          <Stack gap={4} w="full">
+            {value.map((listItemValue, i) => (
+              <Flex alignItems={'center'} gap={2.5}>
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: add key prop

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+                <Argument
+                  name={`${name}.${i}`}
+                  type={listItemsType}
+                  handleChange={handleChange}
+                  error={error}
+                  value={listItemValue}
+                />
+                <Icon h={3} w={3} mt={6} style={{ cursor: 'pointer' }} onClick={() => remove(i)}>
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: can we use iconbutton here since it's clickable?
could use cursor="pointer" directly as attr without style prop
needs area-label

src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+  CommonArgumentInputProps & {
+    type: ClarityAbiTypeTuple;
+    value: TupleValueType;
+    tuple: ClarityAbiTypeTuple['tuple'];
+  }
+> = ({ name, type, handleChange, error, value, tuple }) => {
+  const isOptional = isClarityAbiOptional(type);
+  const tupleType = isOptional ? (type.optional as ClarityAbiTypeTuple) : type;
+  return (
+    <HStack id={name} gap={4} width="100%">
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx]: Field.Root

src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <Stack gap={2} w="full">
+            <Field.Label textStyle="text-regular-sm" color={error ? 'textError' : 'textSecondary'}>
+              {`${name}: ${tupleEntry.name}${isOptional ? ' (optional)' : ''}`}
+            </Field.Label>
+            <Input
+              variant="redesignPrimary"
+              width="full"
+              type={getTypeString(tupleType.tuple[i].type).includes('int') ? 'number' : 'text'}
+              name={`${name}.${tupleEntry.name}`}
+              id={name}
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx]: this id will be the same for elements, maybe use name+tupleEntry.name

src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+              onChange={handleChange}
+              value={value[tupleEntry.name]}
+              placeholder={`${getTypeString(tupleType.tuple[i].type)}`}
+            />
+            <Field.ErrorText color={'textError'} textStyle="text-medium-xs">
+              {error}
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx]: this looks to be an error on tuple level rather than field level, if that's the case we should move it up outside of the map

src/app/txid/[txId]/redesign/function-called/Argument/types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export interface ArgumentInputProps {
+  type: ClarityAbiType;
+  value: ValueType | ListValueType;
+}
+
+export interface CommonArgumentInputProps {
+  name: string;
+  handleChange: (e: React.FormEvent<HTMLInputElement>) => void;
+  error: any;
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/Argument/types.ts]: can we eliminate the `any`?

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
```diff
@@ -0,0 +1,324 @@
...
+  handleCancel?: () => void;
+}
+
+interface FunctionParameters {
+  [key: string]: ValueType | ListValueType;
+}
+
+export type FormikSetFieldValueFunction = (
+  field: string,
+  value: any,
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: can we remove any?

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
```diff
@@ -0,0 +1,324 @@
...
+      }
+      validateOnChange={false}
+      validateOnBlur={false}
+      validate={values => {
+        const functionParametersErrors = checkFunctionParameters(fnAbi, values);
+        const postConditionParametersErrors = checkPostConditionParameters(values);
+        const errors = Object.assign({}, functionParametersErrors, postConditionParametersErrors);
+        return errors;
+      }}
+      onSubmit={async values => {
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx:114]: this function is quite large, can we extract it out of the component and add a unit test?

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
```diff
@@ -0,0 +1,324 @@
...
+            return;
+          }
+          const type = fnAbi.args.find(({ name }) => name === arg)?.type;
+          if (!type) return;
+          const tuple = getTuple(type);
+          const isList = isClarityAbiList(type);
+          const optionalType = isClarityAbiOptional(type) ? type?.optional : undefined;
+          if (tuple) {
+            if (optionalType) {
+              final[arg] = encodeOptionalTuple(tuple, values[arg] as TupleValueType);
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: this function has a lot of unsafe type casting, can we create a few type guard functions e.g. isTupleValueType and put error handling/logging to make sure we capture runtime bugs if they ever happen?

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
```diff
@@ -0,0 +1,324 @@
...
+          postConditionAddress,
+          postConditionConditionCode,
+          postConditionAmount,
+          postConditionAssetAddress,
+          postConditionAssetContractName,
+          postConditionAssetName,
+        } = values;
+
+        if (fnAbi.access === 'public') {
+          void openContractCall({
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: add error handling in case openContractCall fails

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
```diff
@@ -0,0 +1,324 @@
...
+            contractAddress: contractId.split('.')[0],
+            contractName: contractId.split('.')[1],
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: create a function to transform contract id + unit test (we might have one already)

src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx
```diff
@@ -121,11 +110,52 @@ export const FunctionCalled = ({
 }: {
   tx: ContractCallTransaction | MempoolContractCallTransaction;
 }) => {
+  const [isFunctionCallOpen, setIsFunctionCallOpen] = useState(false);
+  const contractCallName = tx.contract_call?.function_name;
+  const contractId = tx.contract_call?.contract_id;
+  const { data: contract } = useContractById(contractId);
+  const functionAbi = contract?.abi?.functions?.find(
+    (fn: any) => fn.name === contractCallName
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx]: can we remove any?

src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx
```diff
@@ -121,11 +110,52 @@ export const FunctionCalled = ({
...
+  ) as unknown as ClarityAbiFunction;
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx:120]: same issue with unsafe type casting

src/app/txid/[txId]/redesign/function-called/FunctionList.tsx
```diff
@@ -0,0 +1,54 @@
...
+    ? functions.filter(fn => fn.name === selectedFunction.name)
+    : functions;
+
+  if (!functions) return null;
+
+  return (
+    <Stack gap={2}>
+      <TabsContentContainer>
+        <Stack gap={4}>
+          {filteredFunctions.map(fn => (
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionList.tsx:27]: filteredFunctions could be undefined

src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+export function FunctionListItem({
+  functionAbi,
+  isOpen,
+  setIsOpen,
+}: {
+  functionAbi: ClarityAbiFunction;
+  isOpen: boolean;
+  setIsOpen: (open: boolean) => void;
+}) {
+  if (!functionAbi) return null;
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx]: seem to be redundant unless FunctionListItem is typed incorrectly

src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+            px={1.5}
+            py={0.5}
+            bg="surfaceFifth"
+            label={<DefaultBadgeLabel label={'Read Only'} fontFamily="instrument" />}
+            icon={<DefaultBadgeIcon icon={<Eye weight="bold" />} color="iconSecondary" />}
+            variant="solid"
+          />
+        )}
+      </Flex>
+      <ButtonLink
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx]: better to use button here instead of buttonlink

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+export function PostConditionForm({
+  values,
+  errors,
+  formikSetFieldValue,
+  handleChange,
+}: {
+  values: FunctionFormikState;
+  errors: FormikErrors<FunctionFormikState>;
+  formikSetFieldValue: FormikSetFieldValueFunction;
+  handleChange: any;
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: can we remove any?

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+    <>
+      {values.postConditionMode === PostConditionMode.Deny && (
+        <Stack gap={4}>
+          <Stack gap={2}>
+            <Select<PostConditionTypeValue, PostConditionTypeLabel>
+              placeholder="Post-condition type"
+              items={PostConditionTypeOptions}
+              onValueChange={details => {
+                const postConditionTypeValue = details
+                  .value[0] as unknown as PostConditionTypeValue;
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: same unsafe type casting and assumption that value[0] exists

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+              </Caption>
+            )}
+          </Stack>
+          {postConditionTypeParameters.length > 0 && (
+            <Stack gap={4}>
+              {postConditionTypeParameters.map(parameter => (
+                <Box key={parameter}>
+                  {parameter !== 'postConditionMode' &&
+                  parameter !== 'postConditionType' &&
+                  parameter !== 'postConditionConditionCode' ? (
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: can you extract this as a boolean with a descriptive name?

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+                      </Field.Root>
+                    </Stack>
+                  ) : (
+                    <Stack gap={2}>
+                      <Select<PostConditionConditionCodeValue, PostConditionConditionCodeLabel>
+                        placeholder="Condition code"
+                        items={postConditionConditionCodeOptions}
+                        onValueChange={details => {
+                          const postConditionConditionCodeValue = details
+                            .value[0] as unknown as PostConditionConditionCodeValue;
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: same issue

src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const { data } = useCallReadOnlyFunction({
+    contractId,
+    fn,
+    readOnlyValue,
+    stacksNetwork: network,
+    stxAddress,
+  });
+
+  const CodeEditorWithControls = useMemo(() => withControls(CodeEditor, true, false), []);
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx]: empty dependency array

src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </Text>
+          <CodeEditorWithControls code={parseReadOnlyResponse(data)} />
+        </Stack>
+      ) : (
+        <Text textStyle="text-regular-sm" color="textPrimary">
+          {data.result}
+        </Text>
+      )}
+      {handleCancel && (
+        <ButtonLink
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx:55]: better to use button

src/app/txid/[txId]/redesign/function-called/post-condition-utils.ts
```diff
@@ -0,0 +1,430 @@
+import { NonTupleValueType } from '@/app/sandbox/types/values';
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/function-called/post-condition-utils.ts]: this file would benefit from unit test

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
```diff
@@ -14,54 +18,131 @@ import { configLanguage } from './editor-config/language';
...
-      if (ref && 'current' in ref) {
-        ref.current = editor;
+export const DEFAULT_EDITOR_HEIGHT = '252px';
+const BUTTONS_HEIGHT = 8;
+
+type CodeEditorProps = {
+  code: string;
+} & Partial<EditorProps>;
+
+const CodeEditorBase = forwardRef<any, CodeEditorProps>(({ code, ...editorProps }, ref) => {
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/source/CodeEditor.tsx:28]: can we remove any?

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
```diff
@@ -14,54 +18,131 @@ import { configLanguage } from './editor-config/language';
...
+    if (claritySyntax) await liftOff(monaco, claritySyntax);
+  }, []);
+  const handleEditorOnMount: OnMount = useCallback(editor => {
+    if (ref && 'current' in ref) {
+      ref.current = editor;
+    }
+    editor.updateOptions({
+      wordSeparators: '`~!@#$%^&*()=+[{]}\\|;:\'",.<>/?',
+    });
+  }, []);
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/source/CodeEditor.tsx]: empty dependency array

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
```diff
@@ -14,54 +18,131 @@ import { configLanguage } from './editor-config/language';
...
+// HOC to add controls to the code editor
+export function withControls(
+  WrappedCodeEditor: typeof CodeEditor,
+  hasCopyButton = true,
+  hasExpandButton = true
+) {
+  return function CodeEditorWithControls({ code, ...editorProps }: CodeEditorProps) {
+    const [codeHeight, setCodeHeight] = useState(DEFAULT_EDITOR_HEIGHT);
+    const [isCodeHeightExpanded, setIsCodeHeightExpanded] = useState(false);
+    const editorRef = useRef<any>(null);
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/source/CodeEditor.tsx:95]: can we remove any?

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
```diff
@@ -14,54 +18,131 @@ import { configLanguage } from './editor-config/language';
...
+          {hasExpandButton && (
+            <Button
+              variant="redesignPrimary"
+              aria-label={'expand source code'}
+              onClick={toggleHeight}
+              p={1.5}
+              h={BUTTONS_HEIGHT}
+              w={BUTTONS_HEIGHT}
+              minW={BUTTONS_HEIGHT}
+              // bg="surfaceInvert"
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/source/CodeEditor.tsx]:
```suggestion
```

src/ui/ButtonLink.tsx
```diff
@@ -11,36 +11,43 @@ import { linkRecipe } from './theme/recipes/LinkRecipe';
...
 
+type ButtonLinkDirection = 'forward' | 'backward';
+
 type ButtonLinkProps = Omit<LinkProps, 'variant' | 'size'> & {
   children: ReactNode;
   buttonLinkSize: ButtonLinkSize;
+  buttonLinkDirection?: ButtonLinkDirection;
 };
 
 const ButtonLinkBase = forwardRef<HTMLAnchorElement, ButtonLinkProps>(
```
He1DAr: do we have this component in design system? we might be good with just Button and Link components.

src/components/ui/select.tsx
```diff
@@ -1,58 +1,166 @@
...
+    const placement = positioning?.placement ?? 'bottom-start';
+    const mainAxis = positioning?.offset?.mainAxis ?? 0;
+
+    return (
+      <Box position="relative">
+        <ChakraSelect.Control
+          {...rest}
+          ref={ref}
+          borderBottomLeftRadius={open ? 'none' : 'redesign.lg'}
+          borderBottomRightRadius={open ? 'none' : 'redesign.lg'}
```
He1DAr, line 29: these are already defined in recipe?

src/components/ui/select.tsx
```diff
@@ -1,58 +1,166 @@
...
+            className="select-tab"
+            position="absolute"
+            height={mainAxis / 4} // dividing by 4 to account for the fact that the popover root's positioning.offset.mainAxis does not seem to abide my chakra ui's normal scaling of input (*4)
+            bottom={-mainAxis / 4}
+            left={0}
+            w="100%"
+            zIndex="docked"
+          >
+            <Icon
+              color="surfacePrimary"
```
He1DAr: can some of these styles be defined in recipe instead?

src/common/utils/sandbox.ts
He1DAr: I'd double check this, it seems to be related to a bug or problem that was reported a while ago.

src/common/components/Select.tsx
```diff
@@ -0,0 +1,96 @@
+import {
```
He1DAr, line 1: what's the difference between using this component or using ui/select directly? if ResizeObserver is needed for all select fields we can maybe add it to ui/select and eliminate this component

src/common/components/CopyButton.tsx
```diff
@@ -16,12 +16,27 @@ export const CopyButton = ({
   );
 };
 
-export type ExtendedIconProps = Omit<IconProps, '_hover'> & {
+export type ExtendedIconProps = Omit<IconProps, '_hover' | '_groupHover' | 'color'> & {
   _hover?: ((copied: boolean) => IconProps['_hover']) | IconProps['_hover'];
+  _groupHover?: ((copied: boolean) => IconProps['_groupHover']) | IconProps['_groupHover'];
```
He1DAr: it looks like this file reuse this pattern of omitting a prop then readding it as a function, you could create a generic type for this e.g. 
type CustomProps<T, K extends keyof T> = Omit<T, K> & {
   [P in K]?: ((copied: boolean) => T[P]) | T[P];
};

src/app/txid/[txId]/redesign/source/Source.tsx
```diff
@@ -34,81 +29,20 @@ export function Source({
...
-    setIsCodeHeightExpanded(!isCodeHeightExpanded);
-    if (isCodeHeightExpanded) {
-      setCodeHeight(DEFAULT_HEIGHT);
-    } else {
-      setCodeHeight(editorRef.current?.getContentHeight());
-    }
-  }, [isCodeHeightExpanded]);
+  const network = useGlobalContext().activeNetwork;
+
+  const CodeEditorWithControls = useMemo(() => withControls(CodeEditor, true, true), []);
```
Comment (He1DAr) [inline src/app/txid/[txId]/redesign/source/Source.tsx]: empty dependency arr

BLuEScioN:
@ginny-d 
Regarding 4
image
The badge used in the figma for the contract name on the contract call tx id page details card and the badge used in the figma for the contract name in the txs table do not match the system design styling for the badge component. I think both of the badges are examples of the simple badge in the design system and thus the styling should be the same. Can you let me know which styling is preferred here?
image

BLuEScioN:
@ginny-d 
Reagrding 13
 "Scroll bug in expanded code editor: When placing the pointer on the code editor, it's not possible to scroll."
 This isn't a scroll bug. You can't scroll in because there is no overflow on the component. The component is fully rendered. The rest of the page is, however, scrollable, so you have to use the outer edge, outside of the code editor to scroll.

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: done

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: done

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: done

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: In my experience, IconButton makes making icon buttons harder than using a button and and icon together. Also, at this point IconButton is a legacy component and is only used in legacy UIs. Since there is no UI component for an icon button, I'd honestly prefer to just phase IconButton out.
Additionally, the figma designs don't have show what this button is supposed to look like. Before I change anything, @ginny-d could you take a look here and see if you want to change anything about this UI
image

src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx]: This does have a Field.Root

src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx]: done

src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/TupleArgumentInput.tsx]: done

ginny-d:
> "Scroll bug in expanded code editor: When placing the pointer on the code editor, it's not possible to scroll."
> This isn't a scroll bug. You can't scroll in because there is no overflow on the component. The component is fully rendered. The rest of the page is, however, scrollable, so you have to use the outer edge, outside of the code editor to scroll.

I think this is still worth addressing because it has UX and accessibility implications (and I’ve already received feedback calling this a “bug”).

When the pointer is over a non-scrollable element, wheel/trackpad scroll events should bubble to the nearest scrollable ancestor. If the rest of the page can scroll, users expect the page to move even if the cursor is positioned over the editor.

A good example of the correct interaction can be seen here: https://etherscan.io/address/0x6aa981bff95edfea36bdae98c26b274ffcafe8d3?utm_source=chatgpt.com#code

src/app/txid/[txId]/redesign/function-called/Argument/types.ts
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/types.ts]: done

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx:114]: done

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCallForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx]: No as this is how functions is typed in @stacks/stacks-blockchain-api-types
image

Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx]: I removed the any though so we can see the real type

src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionCalled.tsx:120]: see above

src/app/txid/[txId]/redesign/function-called/FunctionList.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionList.tsx:27]: Maybe I'm missing something but If there isnt a selected function filteredfunctions will just be functions. If there is a selected function, the selected function has to be one of the clarityabifunctions in the functions we received from the abi.

src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx]: done

src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/FunctionListItem.tsx]: It is a button, but we have this buttonlink component that we already built and use throughout the app that accomplishes what is shown in the designs. I am extending the buttonlink component to be able to be turned into a button instead of a link

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: Unfortunately, ChakraUI statically types details.value as a string[]. There is no way to pass type parameters to change it. I have to cast details.value[0] in order to index into PostConditionTypeValueMapReversed. I am using this map because I need to be able to set the formik field state to a PostConditionType from @stacks/transactions as that value will be used in openContractCall to create the post condition. I could not use PostConditionType directly as a value in the select options because it is an enum without a label, so it's type is number, not string.

We expect details.value to always be a string array. The select is always expected to hold at least one value. We can add a check here but it's not necessary imo

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/PostConditionForm.tsx]: done

src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx]: fixed. put outside compoennt

src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/ReadOnlyField.tsx:55]: see previous comment

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/source/CodeEditor.tsx:28]: done

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/source/CodeEditor.tsx]: done

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/source/CodeEditor.tsx:95]: done

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/source/CodeEditor.tsx]: done

src/ui/ButtonLink.tsx
BLuEScioN: yes
here https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=1-9&p=f&t=Lxh4jMmDvjNf4b4y-0
image

src/app/txid/[txId]/redesign/source/CodeEditor.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/source/CodeEditor.tsx:28]: actually it isn't possible to get the correct type editor.IStandaloneCodeEditor as it isn't exported. Since the type is complex I am using any

src/components/ui/select.tsx
BLuEScioN: Dyanmic styles that rely on state cannot be put in recipes

src/components/ui/select.tsx
BLuEScioN: These cannot as they are dynamic components that are not a part of the default slot components for select

src/common/components/Select.tsx
BLuEScioN: ui/select is the file created by using the add snippet command. This file has all the composable components that make up a select control. Sure you could piece together the components as you like for each instance you use a select, but that would be tedious. The select component here (src/common/components/Select.tsx) is how I want to use the composable select components and unify all the props into one parent select component. It also has the extra logic needed to make tabs work with the select control.

src/common/components/CopyButton.tsx
BLuEScioN: I see. But I am not sure this makes the code any simpler or easy to understand

src/app/txid/[txId]/redesign/source/Source.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/source/Source.tsx]: fixed

src/app/txid/[txId]/redesign/function-called/post-condition-utils.ts
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/post-condition-utils.ts]: done

src/common/components/CopyButton.tsx
He1DAr: same complexity, will just make this part DRY

ginny-d:
> In my experience, IconButton makes making icon buttons harder than using a button and and icon together. Also, at this point IconButton is a legacy component and is only used in legacy UIs. Since there is no UI component for an icon button, I'd honestly prefer to just phase IconButton out.
> Additionally, the figma designs don't have show what this button is supposed to look like. Before I change anything, @ginny-d could you take a look here and see if you want to change anything about this UI

@BLuEScioN  Apologies but I didn't understand if we're talking about the "+ Add item" button here or the "x" button?
If it's the "x", we could use the same styling as the "copy" button and make that a component for "icon button"?

ginny-d:
@BLuEScioN thanks, here's my last revision!

- [x] Font for empty state message: Currently "No post-conditions to show" looks to big in comparison with the rest of the page
Proposal: Use text-regular-sm with text-tertiary
Screenshot 2025-09-04 at 17 34 12

-----------
- [x] Dropdown component mismatch
Current: "Allow mode" / "Deny mode" dropdown uses a different component than filters.
Expected: Use the same component as the filter dropdown for consistency.

Screenshot 2025-09-04 at 17 34 49

-----------
- [x] Events table column headers alignment
Current: Post-condition table columns (From / Condition / Asset / To) are misaligned.
Expected: Proper spacing and column alignment.

Screenshot 2025-09-04 at 17 35 14

-----------
- [x] Here we are missing the Clarity icon, to keep it consistent with the tag shown on tables
Screenshot 2025-09-04 at 17 35 24

-----------

- [x] Cancel button:
Current: "Cancel" button uses "warning" styling. As this is not a warning button but a simply "cancel action", red is not appropriate
Expected: Use button-sm, tertiary variant (outlined), to reduce emphasis.

Screenshot 2025-09-04 at 17 36 14


-------------

- [x] Mobile: Post-condition table causes overflow, and header is not properly aligned to the left

Screenshot 2025-09-04 at 17 37 00

---------

- [x] Tag components (Success / Allow mode / Deny mode)
Current: Inconsistent styling and spacing between tags (icon alignment, padding).
Expected: All tags should use the same component and spacing rules. 
Screenshot 2025-09-04 at 17 44 20

---------

- [x] If the table has only one line, can the column header (in this case, Result), be aligned vertically?

Screenshot 2025-09-04 at 17 39 19

src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx
Comment (BLuEScioN) [reply src/app/txid/[txId]/redesign/function-called/Argument/ListArgumentInput.tsx]: @ginny-d Can you respond here please?

BLuEScioN:
> > In my experience, IconButton makes making icon buttons harder than using a button and and icon together. Also, at this point IconButton is a legacy component and is only used in legacy UIs. Since there is no UI component for an icon button, I'd honestly prefer to just phase IconButton out.
> > Additionally, the figma designs don't have show what this button is supposed to look like. Before I change anything, @ginny-d could you take a look here and see if you want to change anything about this UI
> 
> @BLuEScioN Apologies but I didn't understand if we're talking about the "+ Add item" button here or the "x" button? If it's the "x", we could use the same styling as the "copy" button and make that a component for "icon button"?

Gotcha. It was the X.

BLuEScioN:
@ginny-d 
"Dropdown component mismatch
Current: "Allow mode" / "Deny mode" dropdown uses a different component than filters.
Expected: Use the same component as the filter dropdown for consistency."

TLDR: I will try to match the styles the best I can, but these are actually 2 separate components and I think applying styles 1:1 will not be totally straightforward. I think it would be best if you could point me to a design file with the exact styles you want. Currently, I was trying to match the styles shown on [tx id page designs ](https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=7347-25733&m=dev) and for the dropdown part, the[ title dropdown designs ](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=2340-24775&m=dev)

These are not the same components under the hood.

There are 2 components here: the popover and the select component.

The popover component was designed off of the examples shown on the [tx page designs ](https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=4321-78035&m=dev)

The select component was designed off the [title dropdown designs](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=2340-24775&m=dev) 

Filters uses the [popover](https://chakra-ui.com/docs/components/popover), which is a bit more of a swiss army knife in the sense that the UI generally calls for a very custom UI to go into the content that pops up when the popover trigger is clicked. For example, all the filter UIs, ie type, date, and from/to, have different UIs inside of them.
  
The [select component](https://chakra-ui.com/docs/components/select) by contrast has much more explicitly defined behavior. It's used to pick a value from predefined options. The advantage of using the select component for the post conditions is that, for the most part, I don't have to add additional logic to get it to do what we want. All the functionality has already been predefined in the design system component itself.

BLuEScioN:
@ginny-d You want button size small for both call function and cancel?
"Cancel button:
Current: "Cancel" button uses "warning" styling. As this is not a warning button but a simply "cancel action", red is not appropriate
Expected: Use button-sm, tertiary variant (outlined), to reduce emphasis."

src/common/utils/sandbox.ts
BLuEScioN: I reverted this

ginny-d:
> @ginny-d You want button size small for both call function and cancel? "Cancel button: Current: "Cancel" button uses "warning" styling. As this is not a warning button but a simply "cancel action", red is not appropriate Expected: Use button-sm, tertiary variant (outlined), to reduce emphasis."

Yes, size small for these buttons

ginny-d:
> @ginny-d "Dropdown component mismatch Current: "Allow mode" / "Deny mode" dropdown uses a different component than filters. Expected: Use the same component as the filter dropdown for consistency."
> 
> TLDR: I will try to match the styles the best I can, but these are actually 2 separate components and I think applying styles 1:1 will not be totally straightforward. I think it would be best if you could point me to a design file with the exact styles you want. Currently, I was trying to match the styles shown on [tx id page designs ](https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=7347-25733&m=dev) and for the dropdown part, the[ title dropdown designs ](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=2340-24775&m=dev)
> 
> These are not the same components under the hood.
> 
> There are 2 components here: the popover and the select component.
> 
> The popover component was designed off of the examples shown on the [tx page designs ](https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=4321-78035&m=dev)
> 
> The select component was designed off the [title dropdown designs](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=2340-24775&m=dev)
> 
> Filters uses the [popover](https://chakra-ui.com/docs/components/popover), which is a bit more of a swiss army knife in the sense that the UI generally calls for a very custom UI to go into the content that pops up when the popover trigger is clicked. For example, all the filter UIs, ie type, date, and from/to, have different UIs inside of them.
> 
> The [select component](https://chakra-ui.com/docs/components/select) by contrast has much more explicitly defined behavior. It's used to pick a value from predefined options. The advantage of using the select component for the post conditions is that, for the most part, I don't have to add additional logic to get it to do what we want. All the functionality has already been predefined in the design system component itself.

I see what you mean. Going forward, I’ll be more explicit about which component variant is being used. The one used here is the [“Single select” dropdown](https://www.figma.com/design/zR7MN402tgacWXm63OQLms/Stacks-Explorer-Design-System--Redesign-?node-id=2322-1898&t=ASO9r1qKcTRCEJdZ-1), while for filters we’d use the “Multi select” option below.

PR #2506, opened 2025-09-10
fix: fixed fn call optional arg validation

BLuEScioN:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Fixed the issue reported [here](https://hiropbc.slack.com/archives/C05T570TNQJ/p1757519082417029) 
The issue was not with our validation. The issue was that, after passing validation, during encoding, we checked if the tuple was optional, but we did not check if the tuple was not optional, if any of its fields were optional

# Checklist:

### DIFF
--- a/src/app/sandbox/utils.ts
+++ b/src/app/sandbox/utils.ts
@@ -38,8 +38,11 @@ export const encodeTuple = (
 ): ClarityValue => {
   const tupleData = tuple.reduce(
     (acc, tupleEntry) => {
-      const _type = tupleEntry.type;
-      acc[tupleEntry.name] = encodeAbiClarityValue(value[tupleEntry.name].toString(), _type);
+      const optionalType = isClarityAbiOptional(tupleEntry.type);
+      const type = tupleEntry.type;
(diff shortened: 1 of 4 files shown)


src/app/sandbox/utils.ts
```diff
@@ -38,8 +38,11 @@ export const encodeTuple = (
 ): ClarityValue => {
   const tupleData = tuple.reduce(
     (acc, tupleEntry) => {
-      const _type = tupleEntry.type;
-      acc[tupleEntry.name] = encodeAbiClarityValue(value[tupleEntry.name].toString(), _type);
+      const optionalType = isClarityAbiOptional(tupleEntry.type);
+      const type = tupleEntry.type;
+      acc[tupleEntry.name] = optionalType
+        ? noneCV()
```
obycode: Does this mean it will only accept `none`? It should be able to be empty, for `none` **or** contain a value, for a `some` value.

BLuEScioN: Yes I should be including the value if its present

PR #2526, opened 2025-09-29
fix: update blocks page pagination

heidar-stacks:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Since we're no longer user "load more" to paginate, we need to move away from infinite query and use https://tanstack.com/query/v4/docs/framework/react/guides/paginated-queries

# Checklist:

### DIFF
--- a/src/app/blocks/queries/useBlocksV2Queries.ts
+++ b/src/app/blocks/queries/useBlocksV2Queries.ts
@@ -11,6 +11,7 @@ import { getNextPageParam } from '../../../common/utils/utils';
 
 export const BLOCKS_V2_INFINITE_QUERY_KEY = 'blocksV2Infinite';
 export const BLOCKS_V2_LIST_QUERY_KEY = 'blocksV2List';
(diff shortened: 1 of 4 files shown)


src/app/blocks/queries/useBlocksV2Queries.ts
```diff
@@ -70,3 +70,24 @@ export const useBlocksV2List = (
...
+export const useBlocksV2Paginated = (
+  pageIndex: number,
+  limit = DEFAULT_LIST_LIMIT,
+  options?: any
+): UseQueryResult<GenericResponseType<NakamotoBlock>> => {
+  const { activeNetwork } = useGlobalContext();
+  const offset = pageIndex * limit;
+
+  return useQuery({
+    queryKey: ['blocksV2Paginated', limit, offset, activeNetwork.url],
```
nick-stacks: make this a function. make blocksV2Paginated a const

PR #2537, opened 2025-10-06
Feat/address id page redesign txs tab

nick-stacks:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds the address id page redesign txs tab

# Checklist:


src/common/queries/useNftMetadata.ts
```diff
@@ -1,20 +1,65 @@
...
+  const isMetadataLoading = nftMetadataQueries.some(query => query.isLoading);
+  const isMetadataFetching = nftMetadataQueries.some(query => query.isFetching);
+  const metadataErrors = nftMetadataQueries.filter(query => query.error).map(query => query.error);
+
+  return {
+    data: nftMetadata,
+    isLoading: isMetadataLoading,
+    isFetching: isMetadataFetching,
+    metadataErrors,
+  };
```
heidar-stacks, line 78: I suggest we split this into 2 hooks, 1 that just returns useQuery similar to other data fetching queries and another one to do the data transformation with unit test

src/common/queries/useFtMetadata.ts
```diff
@@ -38,3 +40,47 @@ export function useSuspenseFtMetadata(
...
+        ...options,
+      })
+    ),
+  });
+
+  // Extract the data from each query result
+  const ftMetadata = ftMetadataQueries.map(query => query.data);
+  const isMetadataLoading = ftMetadataQueries.some(query => query.isLoading);
+  const isMetadataFetching = ftMetadataQueries.some(query => query.isFetching);
+  const metadataErrors = ftMetadataQueries.filter(query => query.error).map(query => query.error);
```
heidar-stacks, line 91: same for this one

src/app/address/[principal]/TokenBalanceCard/TokenAvatar.tsx
```diff
@@ -1,11 +1,9 @@
...
 
-import React from 'react';
-
 import { DefaultTokenImage } from './DefaultTokenImage';
 import { TokenImage } from './TokenImage';
 import { TokenVideo } from './TokenVideo';
-import { useImageUrl } from './useImageUrl';
+import { useImageContentType } from './useImageUrl';
 
 export function TokenAvatar({
```
Comment (heidar-stacks) [inline src/app/address/[principal]/TokenBalanceCard/TokenAvatar.tsx:8]: Since this component is used inside TokenRow, which is rendered in a loop, it will trigger one network request per row which is quite problematic. We'll likely need a backend update to avoid this.

src/common/queries/useNftMetadata.ts
nick-stacks: This isn't doing any data transformations. It's only extracting the data into a convenient format. I am splitting this and writing tests, but I think this is overkill.

The original functions were straightforward wrappers around useQueries with basic data transformation
The "transformation" is just mapping over query results - not complex business logic.
Adding an extra abstraction layer for simple array operations may not provide meaningful value as most consumers will want the transformed data anyway.
The real logic being tested is in React Query itself, not our thin wrapper.
I feel like this is just adding more files to maintain (additional test files)

Before requesting additional tests or refactoring for simple utility hooks, could we consider the cost vs. value.

Are we testing trivial logic (basic array operations) or meaningful business logic?
Will the new abstractions actually be reused, or are we creating unused exports?
Does the complexity added justify the maintenance overhead?

src/app/address/[principal]/TokenBalanceCard/TokenAvatar.tsx
Comment (nick-stacks) [reply src/app/address/[principal]/TokenBalanceCard/TokenAvatar.tsx:8]: Yes, this is how prod works today. The only way to fix these things is, as you proposed, improving the backend.

src/common/queries/useNftMetadata.ts
heidar-stacks: I get your point and agree the transform here is trivial. My concern is less about complexity and more about consistency and testability. We don't unit test useQuery-wrapped hooks since we know useQuery itself is already tested, but once we start operating on the response I think unit tests add value. Separating raw vs derived layers makes the code cleaner and gives us a clean structure for tests, which I think will pay off in maintainability long term

src/app/address/[principal]/TokenBalanceCard/TokenAvatar.tsx
Comment (heidar-stacks) [reply src/app/address/[principal]/TokenBalanceCard/TokenAvatar.tsx:8]: since it's already in prod we can tackle this later https://github.com/stx-labs/explorer/issues/2544

PR #2543, opened 2025-10-09
Feat/token id page scaffolding

nick-stacks:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

# Checklist:
- [ ] I have performed a self-review of my code.

### DIFF
--- a/src/app/token/[tokenId]/page-data.ts
+++ b/src/app/token/[tokenId]/page-data.ts
@@ -1,0 +1,215 @@
+import { fetchTokenDataFromLunarCrush, fetchTokenMetadata } from '@/api/data-fetchers';
+import { logError } from '@/common/utils/error-utils';
+
+import { getFtDecimalAdjustedBalance } from '../../../common/utils/utils';
+import { getIsSBTC } from '../../tokens/utils';
+import {
+  DeveloperDataRedesign,
+  MergedTokenData,
+  TokenDataFromLunarCrush,
+  TokenDataFromStacksApi,
+  TokenLinks,
+} from './types';
+
+async function getTokenDataFromStacksApi(
+  tokenId: string,
+  apiUrl: string
+): Promise<TokenDataFromStacksApi | undefined> {
+  try {
+    const tokenMetadata = await fetchTokenMetadata(apiUrl, tokenId);
+
+    const name = safeGet(tokenMetadata?.name, tokenMetadata?.metadata?.name);
+    const symbol = tokenMetadata?.symbol;
+    const decimals = tokenMetadata?.decimals;
+    const totalSupply = tokenMetadata?.total_supply;
+    const imageUri = tokenMetadata?.image_uri;
+
+    if (!name || !symbol) {
+      throw new Error(`Token not found. tokenId: ${tokenId}, apiUrl: ${apiUrl}`);
+    }
+
+    return {
+      name,
+      symbol,
+      totalSupply:
+        totalSupply && decimals ? getFtDecimalAdjustedBalance(totalSupply, decimals) : undefined,
+      decimals,
+      imageUri,
+    };
+  } catch (error) {
+    logError(error as Error, 'getTokenInfoFromStacksApi', { tokenId, apiUrl }, 'error');
+    return undefined;
+  }
+}
+
+async function getTokenDataFromLunarCrush(
+  tokenId: string
+): Promise<TokenDataFromLunarCrush | undefined> {
+  try {
+    const tokenDataResponse = await fetchTokenDataFromLunarCrush(tokenId);
+
+    const name = tokenDataResponse?.data?.name;
+    const symbol = tokenDataResponse?.data?.symbol;
+
+    const categories: string[] = [];
+
+    const circulatingSupply = tokenDataResponse?.data?.circulating_supply;
+
+    const currentPrice = tokenDataResponse?.data?.price;
+    const currentPriceInBtc = tokenDataResponse?.data?.price_btc;
+    const priceChangePercentage24h = tokenDataResponse?.data?.percent_change_24h;
+    const priceInBtcChangePercentage24h = undefined; // TODO: Why do we need this?
+
+    const marketCap = tokenDataResponse?.data?.market_cap;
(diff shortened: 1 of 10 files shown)


src/app/token/[tokenId]/page-data.ts
```diff
@@ -0,0 +1,256 @@
...
+import { HolderResponseType } from './Tabs/data/useHolders';
+import {
+  DeveloperDataRedesign,
+  MergedTokenData,
+  TokenDataFromLunarCrush,
+  TokenDataFromStacksApi,
+  TokenLinks,
+} from './types';
+
+async function getCirculatingSupplyFromHoldersEndpoint(apiUrl: string, tokenId: string) {
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/page-data.ts]: few comments for data fetching:
1) other files use stacksAPIFetch, we should probably use it for consistency and to be able to change headers easily later
2) we might wanna rethink using console.error() vs logError() or have clear idea when to use which
3) consider adding revalidation/tags for caching

src/app/token/[tokenId]/page.tsx
```diff
@@ -1,24 +1,151 @@
...
+  const isRedesign = searchParams.redesign === 'true';
+  const isSSRDisabled = searchParams?.ssr === 'false';
+
+  if (isRedesign && !isSSRDisabled) {
+    tokenData = await getTokenDataRedesign(tokenId, apiUrl, !!api);
+    try {
+      const [tokenPriceResult, tokenDataResult, contractInfoResult, recentTransactionsResult] =
+        await Promise.allSettled([
+          getTokenPrice(),
+          getTokenDataRedesign(tokenId, apiUrl, !!api),
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/page.tsx]: why we need to call getTokenDataRedesign twice here?

src/app/token/[tokenId]/page.tsx
```diff
@@ -1,24 +1,151 @@
...
+      const abi: ContractInterfaceResponse = contractInfo
+        ? JSON.parse(contractInfo?.abi)
+        : undefined;
+      const ftName = abi ? abi.fungible_tokens[0].name : undefined;
+      assetId = ftName ? `${tokenId}::${ftName}` : undefined;
+
+      const [txResult, holdersResult] = await Promise.allSettled([
+        // dependent queries
+        fetchTx(apiUrl, txId || ''),
+        ...(assetId ? [fetchHolders(apiUrl, assetId, 10, 0)] : []),
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/page.tsx]: if assetId is undefined, wouldn't that make holdersResult undefined and cause null pointer exception in handleSettledResult?

src/app/token/[tokenId]/page.tsx
```diff
@@ -1,24 +1,151 @@
...
+        'error'
+      );
+    }
+  }
+
+  return (
+    <TokenIdPageDataProvider
+      tokenId={tokenId}
+      tokenData={tokenData}
+      redesignTokenData={redesignTokenData}
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/page.tsx:136]: this variable seems to never get assigned a value

src/app/token/[tokenId]/page.tsx
```diff
@@ -1,24 +1,151 @@
...
+      const recentAddressTransactions = handleSettledResult(
+        recentTransactionsResult,
+        'Failed to fetch recent transactions'
+      );
+      const contractInfo = handleSettledResult(contractInfoResult, 'Failed to fetch contract info');
+
+      // tokenPrice = await getTokenPrice();
+      // tokenData = await getTokenInfoRedesign(tokenId, apiUrl, !!api);
+      // const contractInfo = await fetchContractInfo(apiUrl, tokenId);
+      // const recentAddressTransactions = await fetchRecentTransactions(apiUrl, tokenId);
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/page.tsx]:
```suggestion
```

src/app/token/[tokenId]/redesign/context/TokenIdPageContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+  tokenId: string;
+  tokenData: MergedTokenData | undefined;
+  redesignTokenData: MergedTokenData | undefined;
+  txBlockTime: number | undefined;
+  txId: string | undefined;
+  assetId: string | undefined;
+  holders: FungibleTokenHolderList | undefined;
+}
+
+const DEFAULT_TOKEN_ID_PAGE_DATA: TokenIdPageDataContextType = {
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/redesign/context/TokenIdPageContext.tsx]: we should be good without default values.

src/app/token/[tokenId]/page.tsx
Comment (nick-stacks) [reply src/app/token/[tokenId]/page.tsx]: removed

src/app/token/[tokenId]/page.tsx
Comment (nick-stacks) [reply src/app/token/[tokenId]/page.tsx]: handled

src/app/token/[tokenId]/page.tsx
Comment (nick-stacks) [reply src/app/token/[tokenId]/page.tsx:136]: thanks for catching that

src/app/token/[tokenId]/page.tsx
Comment (nick-stacks) [reply src/app/token/[tokenId]/page.tsx]: removed

src/app/token/[tokenId]/page-data.ts
Comment (nick-stacks) [reply src/app/token/[tokenId]/page-data.ts]: I am just now realizing I am already fetching holders separately anyways so I'm going to remove this from the token data from stacks api call/logic

src/app/token/[tokenId]/redesign/context/TokenIdPageContext.tsx
Comment (nick-stacks) [reply src/app/token/[tokenId]/redesign/context/TokenIdPageContext.tsx]: Not sure what you mean here. I use this to create the context:

```
const TokenIdPageDataContext = createContext<TokenIdPageDataContextType>(
  DEFAULT_TOKEN_ID_PAGE_DATA
);
```

Comment (heidar-stacks) [reply src/app/token/[tokenId]/redesign/context/TokenIdPageContext.tsx]: If we have the guard condition to not allow using the hook without provider we should be able to remove the default values, similar to `src\app\_components\BlockList\BlockListContext.ts`

Comment (nick-stacks) [reply src/app/token/[tokenId]/redesign/context/TokenIdPageContext.tsx]: done

PR #2559, opened 2025-10-20
Fix/events table print rows

nick-stacks:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
adds table sub rows that enable us to add the value of print events logs underneath the event row.
this PR also fixes an issue with pagination on the events table that is being caused by an API bug where the total property is not being returned on a list endpoingt
Issue created on the API for the bug - https://github.com/hirosystems/stacks-blockchain-api/issues/2362

https://github.com/stx-labs/explorer/issues/2554

# Checklist:


src/app/txid/[txId]/redesign/events-table/EventsTable.tsx
```diff
@@ -268,6 +284,20 @@ export function EventsTable({
...
 }
+
+function renderSubComponent({ row }: { row: Row<EventsTableData> }) {
+  const printValue = row.original[EventsTableColumns.PrintValue];
+  const formattedPrintValue = printValue ? handleContractLogHex(printValue) : undefined;
+  if (!formattedPrintValue) return null;
+  return (
+    <pre style={{ fontFamily: 'matterMono', fontSize: 'sm' }}>
+      <code>{JSON.stringify(formattedPrintValue, null, 2)}</code>
+    </pre>
```
Comment (heidar-stacks) [inline src/app/txid/[txId]/redesign/events-table/EventsTable.tsx]: Can you format the log so it's more readable? something like:
```suggestion
    <Box as="pre" fontFamily="matterMono" fontSize="sm" whiteSpace="pre">
      <Code display="block" p={3}>
        {formattedPrintValue}
      </Code>
    </Box>
```

src/app/txid/[txId]/redesign/events-table/EventsTable.tsx
```diff
@@ -189,6 +198,10 @@ export function EventsTable({
     isCacheSetWithInitialData.current = true;
   }
 
+  // There is a bug in the API where the tx events endpoint does not return total (total # events for the tx), which is causing pagination to not work correctly. So, we have to fetch the tx to get the total number of events
+  const { data: tx } = useTxById(txId);
```
Comment (heidar-stacks) [inline src/app/txid/[txId]/redesign/events-table/EventsTable.tsx:202]: I think this would lose a big part of the benefit of doing SSR. Could you use the value from the tx  fetched on SSR instead?

src/app/txid/[txId]/redesign/events-table/EventsTable.tsx
Comment (nick-stacks) [reply src/app/txid/[txId]/redesign/events-table/EventsTable.tsx]: done

src/app/txid/[txId]/redesign/events-table/EventsTable.tsx
Comment (nick-stacks) [reply src/app/txid/[txId]/redesign/events-table/EventsTable.tsx:202]: Events table is, so far, always hidden behind a tab. It never uses SSR data

Comment (nick-stacks) [reply src/app/txid/[txId]/redesign/events-table/EventsTable.tsx:202]: I created a task on the API to fix this. My hope is that this hack is temporary :/

Comment (heidar-stacks) [reply src/app/txid/[txId]/redesign/events-table/EventsTable.tsx:202]: @nick-stacks I think the request is happening for all page visits, before expanding the events tab

Comment (nick-stacks) [reply src/app/txid/[txId]/redesign/events-table/EventsTable.tsx:202]: I just pushed the fix. The tabs needed to be lazymounted

heidar-stacks (approved):

PR #2643, opened 2025-11-20
Feat/sbtc token id page

nick-stacks:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds the sbtc token id page header

# Checklist:


src/app/token/[tokenId]/PageClient.tsx
```diff
@@ -1,11 +1,53 @@
...
+import { TokenStats } from './redesign/TokenStats';
+import { TotalSupplyCard } from './redesign/TotalSupplyCard';
 
 export default function TokenIdPageRedesign() {
+  const { tokenId } = useTokenIdPageData();
+  const isSbtc = isSBTC(tokenId);
+
+  if (isSbtc) {
+    return (
+      <Stack gap={8}>
```
Comment (heidar-stacks) [inline src/app/token/[tokenId]/PageClient.tsx]: please refactor this out as a separate component

ginny-stacks:
Here's the QA review:

- [x] **1) Update bridge links:**
Update “Get sBTC” button and “How can I get sBTC?” links to point to https://sbtc.stacks.co/ (this was changed recently).

----------------------------------

- [x] **2) Correct URL:**
Confirm that this token page should be accessible at: `explorer.stacks.co/sbtc`

----------------------------------

- [x] **3) Color space (P3):**
Confirm that colors use the P3 color profile in CSS to maintain Stacks’ brand color intended vibrancy.
See for Reference: https://webkit.org/blog/10042/wide-gamut-color-in-css-with-display-p3/

Implementation:
h

Figma:
i

----------------------------------

- [x] **4) Responsive market data boxes:**
Market data boxes should adapt to the viewport width.

Implementation:


Figma:


----------------------------------

- [x] **5) USD formatting issue (site-wide):**
When prices are shown in USD, the text currently appends “US” before the numbers, making them cumbersome and harder to read. Remove the “US” prefix entirely and ensure the $ symbol uses the standard single-stroke glyph.

Implementation:


Figma:


----------------------------------

- [x] **6) The bottom-right corner of the “Total Supply”** box should use a 30px radius to better match the circular token illustration.




----------------------------------

- [x] **7) sBTC icon alignment:**
The sBTC icon appears visually off-center within its circle. Use the optimized version attached in the issue.

Implementation:


Figma:



**(Right click and save here ⬇️)**
sbtc-tokenicon


----------------------------------

- [x] **8) Remove redundant market data block:**
The market data block on this page is redundant with the header. Replace it with links to the protocol contracts.






----------------------------------

- [x] **9) FAQ card interactions & design:**
FAQ cards are missing the pointer cursor on hover to indicate interactivity.
Design revision: adding a gradient overlay to indicate that the content continues (to avoid the appearance of being cropped).
See: https://www.figma.com/design/gsgfw9WmMbETBEC4FbY2H9/Stacks-Explorer--Redesign-?node-id=9100-37626&t=49uHOLjl8cgDy2do-1

a
b


----------------------------------

- [x] **10) Market data boxes, font & USD issue:**
The “US” issue appears again here. Also please confirm that font styles use `heading-sm` as intended.

Implementation:
e

Figma:
f

----------------------------------

- [x] **11) Remove SIP-10 data banner (this page only):**
For this sBTC token page, remove the SIP-10 compliance banner. It may create unnecessary concern; we can guarantee its accuracy internally as it is our own product.

k

----------------------------------

- [x] **12) Mobile sticky header icon size:**
The token image in the mobile sticky header is too large. It should scale to 24×24px.

Implementation:
m


Figma:
n
o

src/app/token/[tokenId]/PageClient.tsx
Comment (nick-stacks) [reply src/app/token/[tokenId]/PageClient.tsx]: done

nick-stacks:
@ginny-stacks Regarding #7, we are actually using the token image we get from the Token Metadata API. I think the optimal solution is to update the token image the API is using vs changing the logic to handle a custom svg just for sbtc.

My message to the API team is [here](https://stackslabs.slack.com/archives/C09FMCJT7SR/p1764006231463739)

ginny-stacks:
> @ginny-stacks Regarding #7, we are actually using the token image we get from the Token Metadata API. I think the optimal solution is to update the token image the API is using vs changing the logic to handle a custom svg just for sbtc.
> 
> My message to the API team is [here](https://stackslabs.slack.com/archives/C09FMCJT7SR/p1764006231463739)

Thanks! 🙏

nick-stacks:
@ginny-stacks I believe I have solved the US$ issue, but I am having a hard time replicating it with even with a VPN. Please confirm if the issue is fixed for you.

Just FYI, the fix is making sure that we are using the narrowsymbol config parameter for the intl.numberformat function
image

ginny-stacks:
Thanks! It looks great, just one last small tweak:

- On the FAQ cards, it would be great to add the interaction to close them upon clicking anywhere outside the box

Screenshot 2025-11-25 at 09 50 11

- Regarding the sBTC token image, the core developers recommended hard-coding the optimized version at this point because updating the image within the contract is non-trivial, which would be the only way that allows the token API to reliably fetch the latest version.


@andresgalante-stacks This should be ready to release after those two updates. The question remains when do we want to release this update, if after the Bridge, or after the updated Withdrawal/Deposit/Transfers tabs, and/or the updated sBTC transaction ID pages. (or just now)

PR #2679, opened 2026-01-05
feat: use system preferred theme

hugo-stacks:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- Support "system" theme
- Use the system theme by default
- Remove some dead code; `ColorModeButton` wasn't used

# Checklist:
- [x] I have tested the change on desktop and mobile (iOS)

Here is a before/after showing the default behavior (no cookie) on a system using dark theme by default.
And the theme changes automatically when users change their browser or os color mode.
You'll notice that the popup is slightly larger because of the addition of a new icon ("monitor") and the label "System" is longer than "Dark" or "Light". I think this larger popup makes sense. And it switches to full width on smaller devices.

Screenshot 2026-01-05 at 18 13 57

### DIFF
--- a/src/app/_components/NewNavBar/ThemeSetting.tsx
+++ b/src/app/_components/NewNavBar/ThemeSetting.tsx
@@ -3,7 +3,7 @@
 import { useColorMode } from '@/components/ui/color-mode';
 import { Text } from '@/ui/Text';
 import { Box, ClientOnly, Flex, Icon } from '@chakra-ui/react';
-import { Moon, SunDim, Monitor } from '@phosphor-icons/react';
+import { Monitor, Moon, SunDim } from '@phosphor-icons/react';
 import { useTheme } from 'next-themes';
 
 import { OptionPicker, OptionPickerOption } from './OptionPicker';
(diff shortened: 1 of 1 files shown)


src/app/_components/NewNavBar/ThemeSetting.tsx
```diff
@@ -28,26 +29,40 @@ const options: OptionPickerOption[] = [
...
+    // A `minWidth` is set so that the size doesn't change when
+    // the label switches between "light" | "dark" | "system"
+    <Flex alignItems="center" justifyContent="space-between" gap={8} minW={220}>
       <Box lineHeight="redesign.short">
         <Text color="textPrimary" fontSize={{ base: 'sm', lg: 'xs' }} fontWeight="medium">
           Theme
         </Text>
         <Text color="textSecondary" fontSize={{ base: 'sm', lg: 'xs' }}>
-          {options.find(option => option.id === colorMode)?.label}
+          {options.find(option => option.id === theme)?.label}
```
hugo-stacks, line 57: We want to show the `theme`, that is `"light" | "dark" | "system"` and not the `resolvedTheme` or `colorMode` that is just `"light" | "dark"`

PR #2837, opened 2026-09-10
feat: add bitcoin staking data and calculations

leah-stacks:
## What type of PR is this?

- [x] Feature

## Description

Adds the data retrieval and calculations for Bitcoin staking: bonds and registrations,
activity and reward history, cycle reward estimates, historical prices, and scheduling helpers.

Includes pagination, event parsing, and reward calculation tests.

Part 2 split from #2827.

### DIFF
--- a/src/app/staking/data.ts
+++ b/src/app/staking/data.ts
@@ -67,6 +67,7 @@ const REVALIDATE_SECONDS = 60;
 export interface BondsPage {
   bonds: Bond[];
   total: number;
+  nextCursor: string | null;
 }
 
 const MAX_PAGE_LIMIT = 50;
@@ -92,6 +93,7 @@ export async function fetchBondsPage(
   return {
     bonds,
     total: data.total ?? bonds.length,
+    nextCursor: data.cursor?.next ?? null,
   };
 }
 
@@ -264,6 +266,7 @@ export async function fetchCycleRewards(
   );
 
   results.forEach(result => {
+    if (result.stakedMicroStx === BigInt(0)) return;
     byCycle[result.cycleNumber] = result;
   });
   return byCycle;
@@ -313,7 +316,8 @@ async function fetchTxsByFunction(
   functionName: string,
   limit: number,
   offset = 0,
-  cache: 'default' | 'no-store' = 'default'
+  cache: 'default' | 'no-store' = 'default',
+  fetchRequest = stacksAPIFetch
 ): Promise<RawTx[]> {
   const transactions: RawTx[] = [];
   while (transactions.length < limit) {
@@ -324,7 +328,7 @@ async function fetchTxsByFunction(
       contract_id: poxContractId,
       function_name: functionName,
     });
-    const response = await stacksAPIFetch(`${apiUrl}/extended/v1/tx?${params}`, {
+    const response = await fetchRequest(`${apiUrl}/extended/v1/tx?${params}`, {
       cache,
       next:
         cache === 'no-store'
@@ -351,7 +355,13 @@ export function readTopic(repr: string): string | undefined {
   return /\(topic "([^"]+)"\)/.exec(repr)?.[1];
 }
 
-async function fetchTxEvents(apiUrl: string, txId: string, settled = false): Promise<string[]> {
+async function fetchTxEvents(
+  apiUrl: string,
+  txId: string,
+  poxContractId: string,
+  settled = false,
+  fetchRequest = stacksAPIFetch
+): Promise<string[]> {
   const eventLimit = 100;
   const reprs: string[] = [];
   let offset = 0;
@@ -360,7 +370,7 @@ async function fetchTxEvents(apiUrl: string, txId: string, settled = false): Pro
       event_limit: String(eventLimit),
       event_offset: String(offset),
     });
-    const response = await stacksAPIFetch(`${apiUrl}/extended/v1/tx/${txId}?${params}`, {
+    const response = await fetchRequest(`${apiUrl}/extended/v1/tx/${txId}?${params}`, {
       cache: 'default',
       next: {
         revalidate: settled ? SETTLED_REVALIDATE_SECONDS : REVALIDATE_SECONDS,
@@ -372,11 +382,12 @@ async function fetchTxEvents(apiUrl: string, txId: string, settled = false): Pro
     }
     const data: {
       event_count?: number;
-      events?: { contract_log?: { value?: { repr?: string } } }[];
+      events?: { contract_log?: { contract_id?: string; value?: { repr?: string } } }[];
     } = await response.json();
     const events = data.events ?? [];
     reprs.push(
       ...events
+        .filter(event => event.contract_log?.contract_id === poxContractId)
         .map(event => event.contract_log?.value?.repr)
         .filter((repr): repr is string => !!repr)
     );
@@ -422,7 +433,7 @@ function describeContractCall(
   fn: string,
   reprs: string[],
   bondsByIndex: Map<number, Bond>
-): { text?: string; bondIndex?: number; amount?: string } {
+): { text?: string; bondIndex?: number; amount?: string; amountUnavailable?: boolean } {
   const find = (topic: string) => reprs.find(repr => readTopic(repr) === topic);
 
   if (fn === 'setup-bond') {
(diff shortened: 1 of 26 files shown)


src/app/staking/reward-estimate.ts
```diff
@@ -0,0 +1,143 @@
...
+  );
+  if (calculationHeight === undefined) return base;
+  const pendingCycle = Math.floor((calculationHeight - firstBurnHeight) / cycleLength);
+  if (pendingCycle !== cycleNumber) return base;
+
+  const active = bonds.filter(
+    bond =>
+      bond.schedule.activation.bitcoin_height < calculationHeight &&
+      bond.schedule.unlock.bitcoin_height >= calculationHeight
+  );
```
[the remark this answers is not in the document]
leah-stacks: Copilot is wrong here. These comparisons match pox-5's approach in `is-bond-active-at-height`.

src/app/staking/reward-metrics.ts
```diff
@@ -0,0 +1,100 @@
...
+      return incomplete;
+    principal += entry.principalSats;
+    rewards += entry.rewardedSats;
+  }
+  if (principal === BigInt(0))
+    return { note: 'No eligible BTC principal was recorded for this bond.' };
+  return {
+    percent: (100 * 50 * Number(rewards)) / Number(principal),
+    note: 'Annualized credited rewards weighted by the BTC eligible in each interval, using the protocol’s 50-interval year. Later withdrawals do not change this rate. Onward payment is separate.',
+  };
```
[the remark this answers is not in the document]
leah-stacks: Will keep this as-is -- the rounding error at these magnitudes is negligible for a percentage displayed to two decimals

src/app/staking/data.ts
```diff
@@ -0,0 +1,792 @@
...
+      const [rewardsPerMicroStx, stakedMicroStx] = await Promise.all([
+        callPoxReadOnly(apiUrl, poxContractId, 'get-rewards-per-token-for-cycle', cycleNumber),
+        callPoxReadOnly(apiUrl, poxContractId, 'get-total-shares-staked-for-cycle', cycleNumber),
+      ]);
+      return { cycleNumber, rewardsPerMicroStx, stakedMicroStx };
+    })
+  );
+
+  results.forEach(result => {
+    byCycle[result.cycleNumber] = result;
```
heidar-stacks, line 267: you could add something like `if (result.stakedMicroStx === BigInt(0)) return` to guard against cycles that has no record

src/app/staking/data.ts
```diff
@@ -0,0 +1,792 @@
...
+      throw new Error(`Failed to fetch transaction ${txId}: ${response.status}`);
+    }
+    const data: {
+      event_count?: number;
+      events?: { contract_log?: { value?: { repr?: string } } }[];
+    } = await response.json();
+    const events = data.events ?? [];
+    reprs.push(
+      ...events
+        .map(event => event.contract_log?.value?.repr)
```
heidar-stacks, line 380: this list can potentially include events from multiple contracts so you may wanna filter by contract_id = poxContractId

src/app/staking/data.ts
```diff
@@ -0,0 +1,792 @@
...
+    const data: { results?: RawTx[] } = await response.json();
+    const rows = data.results ?? [];
+    transactions.push(...rows);
+    if (rows.length < pageLimit) break;
+  }
+  return transactions;
+}
+
+export function readUint(repr: string, key: string): bigint | undefined {
+  const match = new RegExp(`\\(${key} u(\\d+)\\)`).exec(repr);
```
heidar-stacks, line 346: regex seem to take the first match, I think it's ok unless we expect a nested tuple, or int which will return undefined

src/app/staking/data.ts
```diff
@@ -0,0 +1,792 @@
...
+
+export async function fetchBondRewards(
+  poxContractId: string,
+  chain: string,
+  api?: string
+): Promise<BondRewards | undefined> {
+  const apiUrl = getApiUrl(chain, api);
+  const txs: RawTx[] = [];
+  const seenTxIds = new Set<string>();
+  for (let offset = 0; ; offset += DISTRIBUTION_TX_PAGE) {
```
heidar-stacks, line 707: I'm a bit worried from this logic since it's unbounded, if you get many pages this can be problematic

src/app/staking/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+export function isBondPending(status: BondStatus): boolean {
+  return status === 'upcoming';
+}
+
+export function formatBtc(sats: bigint, decimals = 4): string {
+  const btc = satsToBtc(sats);
+  if (btc === 0) return '0 BTC';
+  if (btc < 0.0001) return `<0.0001 BTC`;
+  const maximumFractionDigits = btc < 1 ? Math.max(decimals, 4) : decimals;
+  return `${btc.toLocaleString(undefined, { maximumFractionDigits })} BTC`;
```
heidar-stacks, line 48: nit: you can use 'en-US' here to lock the locale

src/app/staking/data.ts
```diff
@@ -0,0 +1,792 @@
...
+      } catch (error) {
+        activityFailure(error as Error);
+        return undefined;
+      }
+    }
+  };
+
+  const bondsByIndex = new Map<number, Bond>();
+  const bondRequests = new Map<number, Promise<Bond | undefined>>();
+  const page = await fetchBondsPage(chain, api).catch(() => undefined);
```
heidar-stacks, line 545: I suggest adding activityFailure here and pass the error

src/app/staking/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+import type { BondStatus } from './data';
+import { burnHeightToApproximateTimestamp } from './projections';
+import { bpsToPercent } from './projections';
+
+export function toBigInt(value: string | undefined | null): bigint {
+  if (!value) return BigInt(0);
+  try {
+    return BigInt(value);
+  } catch {
+    return BigInt(0);
```
heidar-stacks, line 14: nit: I prefer returning undefined here and handle it as amountUnavailable vs returning 0

src/app/staking/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  const stx = microStxToStx(microStx);
+  return `${stx.toLocaleString(undefined, { maximumFractionDigits: decimals })} STX`;
+}
+
+export function formatRatePercent(bps: number, decimals = 2): string {
+  return `${bpsToPercent(bps).toFixed(decimals)}%`;
+}
+
+export function formatUsd(amount: number): string {
+  if (!Number.isFinite(amount)) return '-';
```
heidar-stacks, line 61: nit: I think other places use N/A so we may wanna follow the same way for consistency

src/app/staking/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  return index === GENESIS_BOND_INDEX ? 'Genesis' : `Bond ${index}`;
+}
+
+const STATUS_LABELS: Record<string, string> = {
+  upcoming: 'Upcoming',
+  active: 'Active',
+};
+
+export function getBondStatusLabel(status: BondStatus): string {
+  return STATUS_LABELS[status] ?? status.charAt(0).toUpperCase() + status.slice(1);
```
heidar-stacks, line 36: nit: guard against type != string or empty string and maybe return "unknown"

src/app/staking/data.ts
```diff
@@ -0,0 +1,792 @@
...
+    next: { revalidate: REVALIDATE_SECONDS, tags: ['staking-bonds'] },
+  });
+  if (!response.ok) {
+    throw new Error(`Failed to fetch bonds: ${response.status}`);
+  }
+  const data: CursorPaginated<Bond> = await response.json();
+  const bonds = [...(data.results ?? [])].sort((a, b) => b.index - a.index);
+  return {
+    bonds,
+    total: data.total ?? bonds.length,
```
heidar-stacks, line 94: should we pass nextCursor here?

src/app/staking/projections.ts
```diff
@@ -0,0 +1,378 @@
...
+    return { satsPerStxPerCycle, satsPerStxPerYear, apyPercent: undefined };
+  }
+
+  const usdRewardedPerStxPerYear = (satsPerStxPerYear / SATS_IN_BTC) * btcPriceUsd;
+  const aprRatio = usdRewardedPerStxPerYear / stxPriceUsd;
+  const cyclesPerYear = getCyclesPerYear(rewardCycleLength);
+  return {
+    satsPerStxPerCycle,
+    satsPerStxPerYear,
+    apyPercent: (Math.pow(1 + aprRatio / cyclesPerYear, cyclesPerYear) - 1) * 100,
```
heidar-stacks, line 97: not sure if we should do this or just display `aprRatio`, worth confirming with product team

heidar-stacks (commented):
LG! left a few comments

src/app/staking/data.ts
leah-stacks: Yeah, good point. I'm going to add some caps here and a timeout so oversized history scans will fail rather than return partial reward totals. This would be a short/medium-term safeguard - mainnet's bond history will be small for awhile. Longer term, we should move aggregation into the API so we don't need to scan transaction history.

I'll reach out to the API team and make sure issues get created & followup.

src/app/staking/projections.ts
leah-stacks: Yeah, I confirmed this calculation with Mitchell. It matches the way stacking-tracker.com does it. Also, I'll make a note to include a tooltip in the UI: "Gross APY assumes this cycle’s return repeats for a year, with rewards reinvested each cycle, before pool or signer fees."

heidar-stacks (approved):

